PR #17, opened 2025-08-01
feat(website): add plugin directory

V3RON:
This pull request adds a plugin directory, which is a list of known plugins created with Rozenite. Users can add their own plugins by modifying the plugin-directory.json file and opening a pull request.


matinzd:
Great move! I was actually thinking of suggesting the same. 🚀

PR #24, opened 2025-08-04
feat(network-activity-plugin): add "copy as cURL" button

NepeinAV:
_Do you accept UI features? This is alpha, so maybe you focused on stabilizing API, not UI_ 

**Main feature:** "Copy as cURL" button in Network Activity Plugin

### What's done:
- Added button to copy HTTP requests as cURL commands
- Support for different data types: JSON, FormData, plain strings with react-native specific stuff in mind. I did't handle file uploading requests because they don't make sense outside of react-native. You can generate cURL, but it will not work anyways :)
- Safe escaping of special characters for running in shell
- Updated NetworkTestScreen screen to verify FormData functionality

### Why:
Developers can quickly copy any network request from DevTools and run it in terminal for debugging. This is the only reason why we still use Reactotron :)

### How it works:
1. Select a request in Network Activity Plugin
2. Click "Copy as cURL" 
3. Paste in terminal and run

**Example result:**
```bash
curl jsonplaceholder.typicode.com/posts \
  -X POST \
  -H 'content-type: application/json' \
  -H 'x-rozenite-test: true' \
  --data-raw '{
        "title": "title",
        "body": "body",
        "userId": 1
  }'
```

### The problems I found:
- `Request` tab can't properly display `multipart/form-data` right now. It is showing just `[Object object]`, so we need to handle it separately

https://github.com/user-attachments/assets/c49939be-308d-4214-91a2-1ac78e036745

### DIFF
--- a/.nx/version-plans/version-plan-1754906598821.md
+++ b/.nx/version-plans/version-plan-1754906598821.md
@@ -1,0 +1,5 @@
+---
+'@rozenite/network-activity-plugin': prerelease
+---
+
+Introduce "Copy as cURL" feature to the Network Activity Plugin, enabling developers to easily copy any HTTP request as a cURL command for debugging purposes in terminal. The implementation includes comprehensive support for different request body types (JSON, FormData, binary data) with proper shell escaping, React Native-specific handling for FormData and Blob objects, and content-type inference.


packages/network-activity-plugin/src/ui/utils/generateCurlCommand.ts
```diff
@@ -0,0 +1,88 @@
...
+
+function isPostRequest(method: string): boolean {
+  return method.toLowerCase() !== 'get';
+}
+
+function addBodyToCurl(curlParts: string[], postData: RequestPostData, headers: Record<string, string>): void {
+  const contentType = getHttpHeaderValue(headers, 'content-type');
+  const dataType = detectPostDataType(postData);
+
+  console.log(postData, contentType, dataType, typeof postData);
```
V3RON, line 52: 👀

V3RON:
> Request tab can't properly display multipart/form-data right now. It is showing just [Object object], so we need to handle it separately

True. The type declarations for both XMLHttpRequest and XHRInterceptor in RN repository are incorrect. The send callback can be called not only with strings, but also with FormData instances and potentially Blobs. Let's make RequestPostData a union of types:

```
type RequestPostData = 
  | { type: 'text', value: string }
  | { type: 'multipart-form', value: Record<string, any> }
  | { type: 'binary' }
```
 
(This is just an example - the exact types may vary)

WDYT?

NepeinAV:
> potentially Blobs

We don't need to handle them separately, because Blob is just an object that look like that:

Снимок экрана 2025-08-05 в 19 22 25

> WDYT?

Maybe there should be two types:

```
export type MultipartFormData = { 
  _parts: [string, string | null][];
};

export type RequestPostData = 
  | { type: 'text', value: string }
  | { type: 'multipart-form', value: MultipartFormData }
  | null
  | undefined;

export type XhrHttpRequestPostData = 
  | string
  | object
  | null 
  | undefined;
```

`RequestPostData` for UI and `XhrHttpRequestPostData` for things like `network-inspector` to keep it raw

So there should be a moment when the request data should be converted from `XhrHttpRequestPostData` to `RequestPostData`. I think the right place is `handleRequestSent`:

```typescript
    const handleRequestSent = (
      data: NetworkActivityEventMap['request-sent']
    ) => {
      const safeStringify = (data: unknown): string => {
        try {
          return typeof data === 'string' ? data : JSON.stringify(data);
        } catch (error) {
          return String(data); 
        }
      };

      const transformPostDataForRequest = (postData: XhrHttpRequestPostData): RequestPostData | undefined => {
        if (!postData) return undefined;

        if (typeof postData === 'object' && '_parts' in postData) {
          return {
            type: 'multipart-form',
            value: postData as MultipartForm,
          };
        }

        return {
          type: 'text',
          value: safeStringify(postData),
        };
      }

      const entry: NetworkEntry = {
        requestId: data.requestId,
        url: data.request.url,
        method: data.request.method,
        headers: data.request.headers,
        postData: transformPostDataForRequest(data.request.postData),
        status: 'pending',
        startTime: data.timestamp,
        type: data.type,
        initiator: data.initiator,
        request: data.request,
      };

      setNetworkEntries((prev) => new Map(prev).set(data.requestId, entry));
    };
```

Then in `RequestTab`:

```typescript
const renderRequestBody = () => {
    if (!selectedRequest?.requestBody) return null;

    const { data } = selectedRequest.requestBody;

    if (!data) {
      return (
        <div className="text-sm text-gray-400">
          No request body for this request
        </div>
      );
    }

    const { type, value } = data;

    if (type === 'text') {
      try {
        const jsonData = JSON.parse(value);
        return (
          <div className="bg-gray-800 p-3 rounded border border-gray-700">
            <JsonTree data={jsonData} />
          </div>
        );
      } catch {
        // Fallback to pre tag if JSON parsing fails
        return (
          <pre className="text-sm font-mono text-gray-300 whitespace-pre-wrap bg-gray-800 p-3 rounded border border-gray-700 overflow-x-auto">
            {value}
          </pre>
        );
      }
    }
    
    if (type === 'multipart-form') {
      // As for temporary solution, we can show a JSON tree
      return (
        <div className="bg-gray-800 p-3 rounded border border-gray-700">
          <JsonTree data={value} />
        </div>
      );
    }
    
    // For unknown types, we can just show the raw value
    return (
      <pre className="text-sm font-mono text-gray-300 whitespace-pre-wrap bg-gray-800 p-3 rounded border border-gray-700 overflow-x-auto">
        {String(value)}
      </pre>
    );
  };
```

And `addBodyToCurl`:

```typescript
function addBodyToCurl(curlParts: string[], postData: RequestPostData): void {
  if (!postData) {
    return;
  }

  if (postData.type === 'multipart-form') {
    const formParts = postData.value._parts.map(([key, value]) => `${key}=${value}`);

    formParts.forEach(part => addCurlParam(curlParts, '--form', escapeShellArg(part)));

    return;
  }

  addCurlParam(curlParts, '--data-raw', escapeShellArg(stringifyData(postData.value)));
}
```

V3RON:
> We don't need to handle them separately, because Blob is just an object that look like that:

That's exactly why we need to handle them separately, as this is not a real response but rather a React Native blob pointer. We should be able to show the user that we received a binary blob of a given type, name, and size. If we leave it as it is now, it will definitely cause confusion at some point.

Generally, the communication protocol (RN <-> devtools) should ideally not rely on any internals. If we leave __parts as an expected property of MultipartFormData, we are effectively binding the protocol to this particular implementation detail. When it changes in the future, we will need to update RN, the protocol, and the devtools UI.

I decided to merge WebSockets in, which caused a lot of conflicts on your side. Sorry about that. I finally had a moment to clean up the implementation a bit, so many files were affected during the process 🙈

NepeinAV:
> as this is not a real response

Are we talking about request or response? I am little bit confused

> We should be able to show the user that we received a binary blob of a given type, name, and size. If we leave it as it is now, it will definitely cause confusion at some point

Yeah, if we want to have separate view for blobs (and `FormData` with blobs) it would be useful. But not in this PR, i guess?

>  should ideally not rely on any internals

I figured out why my `instanceof` checks didn't work. It's because of serialization between DevTools and RN 🫠. Yeah, that's better not to rely on specific internal thing. 

I did a little bit of research and found what RN accepts in fetch `body` and how conversion working:

```js
// react-native/Libraries/Network

'use strict';

const Blob = require('../Blob/Blob');
const binaryToBase64 = require('../Utilities/binaryToBase64');
const FormData = require('./FormData');

export type RequestBody =
  | string
  | Blob
  | FormData
  | {uri: string, ...}
  | ArrayBuffer
  | $ArrayBufferView;

function convertRequestBody(body: RequestBody): Object {
  if (typeof body === 'string') {
    return {string: body};
  }
  if (body instanceof Blob) {
    return {blob: body.data};
  }
  if (body instanceof FormData) {
    return {formData: body.getParts()};
  }
  if (body instanceof ArrayBuffer || ArrayBuffer.isView(body)) {
    /* $FlowFixMe[incompatible-call] : no way to assert that 'body' is indeed
     * an ArrayBufferView */
    return {base64: binaryToBase64(body)};
  }
  return body;
}

module.exports = convertRequestBody;

```

So final types may look like this, and conversion should be **before** sending data to DevTools (in `handleRequestSend` i think)

```typescript
export type XhrHttpRequestPostData = 
  | string
  | Blob
  | FormData
  | object
  | ArrayBuffer 
  | null 
  | undefined;

export type RequestPostData = 
  | { type: 'text', value: string }
  | { type: 'form-data', value: Record<string, string | object> } // string | object — all values should be type of `string` or `{uri: string, ...}`
  | { 
      // Blob and ArrayBuffer
      type: 'binary', 
      // we don't need the actual binary data right now, because serialization may slow down RN app
      value: {
        size: number;
        type?: string;
        name?: string;
      } 
    }
  | null
  | undefined;
```

What do you think?

V3RON:
> Are we talking about request or response? I am little confused

Right, request! Sorry.

I'm not a fan of the 'object' type. I prefer to use 'unknown', but it may cause issues along the way. Could you try replacing it with 'unknown', and if it's not a good fit in this case, fall back to 'object'? Other than that, it looks great 🔥

NepeinAV:
> Could you try replacing it with 'unknown', and if it's not a good fit in this case, fall back to 'object'?

Used `unknown`, seems like no issues.

Do you think we need to do inferring of `Content-Type` based on body type? For example, when sending `FormData`, we don't need to manually set the `Content-Type` header. However, in this case, the `Content-Type` actually will not be `text/plain` as it shows in DevTools right now. And i believe this applies on binary too

V3RON:
> Do you think we need to do inferring of `Content-Type` based on body type? For example, when sending `FormData`, we don't need to manually set the `Content-Type` header. However, in this case, the `Content-Type` actually will not be `text/plain` as it shows in DevTools right now. And i believe this applies on binary too

This is only true for `FormData`:
https://github.com/facebook/react-native/blob/de5093c88771977b58f7bec3f3ffa64a9595334e/packages/react-native/Libraries/Network/RCTNetworking.mm#L345-L349

We should mimic the same logic as native layer doesn't inform JS thread about final headers 😞

NepeinAV:
> We should mimic the same logic as native layer doesn't inform JS thread about final headers 😞

Pushed this

```
[Start]
   |
   v
[Does headers contain Content-Type?]
   |         \
  Yes        No
   |           \
[Return]   [Can infer Content-Type from postData?]
 headers         |         \
             No         Yes
              |           \
         [Return]   [Add Content-Type]
          headers      to headers
                           |
                           v
                    [Return updated headers]
```

If everything is looking good i will also push `version plan` for this feature 🚀

V3RON (approved):

V3RON:
@NepeinAV Version plan and we are good to go 🎉

NepeinAV:
Done 🚀

PR #30, opened 2025-08-05
feat(runtime): add welcome screen

### LINKED ISSUE #28: Add a welcome screen when Rozenite loads successfully
### Description

Currently, it’s not immediately clear if Rozenite has loaded successfully. Users need to open Chrome DevTools to see logs indicating that everything is working properly.

### Suggested solution

Create a welcome screen that will be displayed in the React Native DevTools, even when no plugins are installed, so it’s clear that Rozenite has loaded successfully.

### Additional context

_No response_

V3RON:
Introduce a welcome screen as the default view for users. This enhancement provides clear visual confirmation that Rozenite has loaded successfully.

Closes #28

### DIFF
--- a/.nx/version-plans/version-plan-1754389663529.md
+++ b/.nx/version-plans/version-plan-1754389663529.md
@@ -1,0 +1,5 @@
+---
+'@rozenite/runtime': prerelease
+---
+
+Introduce a welcome screen as the default view for users. This enhancement provides clear visual confirmation that Rozenite has loaded successfully.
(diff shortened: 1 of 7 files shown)


ferretwithaberet:
> This enhancement provides clear visual confirmation that Rozenite has loaded successfully.

This is nice, and right, but there should be an option to hide it once you confirm it works. I could not find any option to hide it from the React Native DevTools settings, so I guess that it needs to become an option on the metro `withRozenite` function.

PR #43, opened 2025-08-07
feat(network-activity-plugin): show host port

NepeinAV:
Added port after hostname, because it is important part of URL

Снимок экрана 2025-08-07 в 03 09 59

### DIFF
--- a/.nx/version-plans/version-plan-1754556947559.md
+++ b/.nx/version-plans/version-plan-1754556947559.md
@@ -1,0 +1,5 @@
+---
+'@rozenite/network-activity-plugin': prerelease
+---
+
+The port is now displayed after the hostname in the URLs. This clarifies the addresses of services running on non-standard ports and eliminates ambiguities when analyzing network activity.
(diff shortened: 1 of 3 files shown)


V3RON:
Good work 👌 
Could you create a version plan for your changes? 

You can create a version plan file with: `pnpm nx release plan prerelease`. When it asks you for a description, you can put just 'lorem' and confirm. It'll generate a file in `.nx` directory, which you can then open with an editor of your liking and put the correct description. It should be a single paragraph only. As for the frontmatter - replace `__default__` with the name of the package, so in your case it will be `'@rozenite/network-activity-plugin'`

Example: https://github.com/callstackincubator/rozenite/blob/main/.nx/version-plans/version-plan-1754476732875.md

NepeinAV:
Done. Do i need this for every feature?

V3RON:
> Done. Do i need this for every feature?

Yup. It's not that important right now, but in the future it'll be used to determine what package needs to be updated and how (patch, minor, major).

PR #48, opened 2025-08-10
fix(network-activity-plugin): handle 204 on response

zmtmaster:
If the user gets 204, No-Content the `.size` will fail.

Issue is that Javascript returned true to the following statement
`typeof null === 'object'` -> `true`

Added some code to handle that case.

### DIFF
--- a/.nx/version-plans/version-plan-1754935600331.md
+++ b/.nx/version-plans/version-plan-1754935600331.md
@@ -1,0 +1,5 @@
+---
+'@rozenite/network-activity-plugin': prerelease
+---
+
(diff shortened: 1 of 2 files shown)


V3RON:
@zmtmaster Could you create a version plan for your changes?

You can create a version plan file with: `pnpm nx release plan prerelease`. When it asks you for a description, you can put just 'lorem' and confirm. It'll generate a file in .nx directory, which you can then open with an editor of your liking and put the correct description. It should be a single paragraph only. As for the frontmatter - replace __default__ with the name of the package, so in your case it will be '@rozenite/network-activity-plugin'

Example: https://github.com/callstackincubator/rozenite/blob/main/.nx/version-plans/version-plan-1754476732875.md

zmtmaster:
Updated the nx file as requested.

PR #72, opened 2025-08-19
fix(network-activity-plugin): robustly handle JSON content types with parameters

aliffazfar:
## Description
This PR fixes an issue encountered while debugging our dinosaur production app, where responses with complex content-type headers (like `application/json,text/plain,*/*`) were incorrectly handled. The fix ensures JSON responses are properly parsed regardless of additional parameters in the content-type header.

## Changes
- Modified JSON detection to use `type.startsWith('application/json')` instead of exact match
- Ensured JSON responses with complex content types are now properly handled
- Maintained all existing content type handlers and fallback behaviors

## Problem Solved
While debugging our dinosaur production app, we discovered:
- Responses with `application/json,text/plain,*/*` content type were incorrectly classified
- These responses fell through to the "other content types" handler
- Were displayed as binary content instead of parsed JSON

Screenshot 2025-08-19 at 2 32 48 PM

## Real-World Validation
Verified with:
- Our dinosaur production app's complex API responses
- Standard `application/json` content type
- Content types with parameters: `application/json; charset=utf-8`
- Complex types: `application/json,text/plain,*/*`
- Other content types: `text/html`, `text/plain`, `application/xml`

## Impact
- Enables proper inspection of JSON responses from real-world production APIs
- Improves network debugging for applications with complex content negotiation


V3RON:
@aliffazfar 
`application/json,text/plain,*/*` is not a valid Content-Type value according to the [HTTP specification](https://www.w3.org/Protocols/rfc1341/4_Content-Type.html). It looks like the value of the `Accept` header instead. Are you sure the server is responding with this as the `Content-Type`? It's more likely that there is a bug in the code mistaking `Accept` for `Content-Type`.

aliffazfar:
@V3RON I believe you are right, I have confirmed that client side are not sending those Content-Type. The bug must be from the server side. But attached here is the behaviour from network tab in chrome, and other network logger library.

Can we consider this approach as a fallback like the other logger behaviour? since we have catch if JSON parsing fails.

Screenshot 2025-08-19 at 7 49 18 PM
Screenshot 2025-08-19 at 8 46 31 PM
Screenshot 2025-08-19 at 8 46 31 PM

V3RON:
If we decide to accept malformed Content-Type headers for application/json, then we should also accept malformed values for others. Please replace all checks in the ResponseTab with .startsWith.

If it works correctly, provide a version plan by running:
pnpm nx release plan prerelease

Follow the format of other version plans in the .nx/version-plans directory.

aliffazfar:
Done @V3RON

PR #90, opened 2025-09-01
proposal: make withRozenite composable

marklawlor:
This is a proposal PR to make `withRozenite` composable - meaning it retains the shape of the `MetroConfig` passed into it and doesn't automatically turn it into a `Promise<MetroConfig>` - which is not accepted by a number of libraries. It also allows passing a function to `withRozenite` as `metro.config.js` which is a valid export of `metro.config.js`

This change would also make `enhanceMetroConfig` redundant, as you could compose the configs together.

For example, if `withRozeniteExpoAtlasPlugin` also implemented this pattern you could compose them together

```js
withRozeniteExpoAtlasPlugin(withRozenite(config))
```

I'm just skimming through the project, so if there is a deeper reason for `enhanceMetroConfig` feel free to close this PR 😄 I created this as a PR instead of an issue to demonstrate how it could be implemented.

### DIFF
--- a/packages/metro/src/index.ts
+++ b/packages/metro/src/index.ts
@@ -16,11 +16,35 @@ export type RozeniteMetroConfig<TMetroConfig = unknown> = Omit<
   ) => Promise<TMetroConfig> | TMetroConfig;
 };
 
-export const withRozenite = async <T extends MetroConfig>(
+export function withRozenite<T extends MetroConfig>(
+  config: () => Promise<T>,
+  options: RozeniteMetroConfig<T>
+): Promise<T>
+export function withRozenite<T extends MetroConfig>(
+  config: () => T,
+  options: RozeniteMetroConfig<T>
+): T
+export function withRozenite<T extends MetroConfig>(
+  config: Promise<T>,
+  options: RozeniteMetroConfig<T>
+): Promise<T>
+export function withRozenite<T extends MetroConfig>(
+  config: T,
+  options: RozeniteMetroConfig<T>
+): T
+export function withRozenite<T extends MetroConfig>(
   config: T | Promise<T>,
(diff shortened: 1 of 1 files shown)


V3RON:
Hi @marklawlor 👋 

> which is not accepted by a number of libraries

Could you provide an example? I'm curious, because I don't immediately see how the Promise returned from the last wrapper would interfere with a library. For instance, that's how Nx and its withNxMetro work. Rozenite should be applied last to ensure it works correctly, since some libraries tend to overwrite the config instead of merging it (usually functions inside the config).

> This change would also make enhanceMetroConfig redundant, as you could compose the configs together.

Its purpose is to prevent plugins from interfering with the config in bundle mode, when creating the final bundle for your application. For example, with Redux DevTools, it prevents the plugin from keeping the process alive indefinitely. Previously, users would wrap their configs multiple times, like this: `withRozenite(withReduxDevTools(config))`, and it still works just fine.

That said, I agree it's best to support as many cases as possible, especially since this "function config" approach is something I wasn't aware of. Applying Rozenite with this form of config could indeed be tricky 🤔

marklawlor:
We had this same issue with Nativewind and were required to make this change. I was surprised how many companies have their own `withCompanyName` wrapper and random libraries that use a `with*` wrapper. We used to document that `withNativewind` should be the last `with*`, but `withNx` also requires this and it looks like `withRozenite` also does - which just leads to confusion for end users.

V3RON:
I'm going to close this pull request. I feel the effort of providing correct type definitions for every possible case (config as an object, config as a Promise, enhanceMetroConfig returning a config, or enhanceMetroConfig returning a Promise) isn't really worth it (unless users show otherwise by opening issues about it 😄 ).

That said, I really appreciate you pointing this out - thank you! 🙌

PR #92, opened 2025-09-02
fix(metro): bundling in Xcode

Jheysoon:
When creating a build to test in TestFlight, i have encountered that it is infinitely in building process.

Debug the issue and found out that isBundling is always false and will setup the metro config with incorrect values

image

### DIFF
--- a/.nx/version-plans/version-plan-1757063298892.md
+++ b/.nx/version-plans/version-plan-1757063298892.md
@@ -1,0 +1,5 @@
+---
+'@rozenite/metro': prerelease
+---
+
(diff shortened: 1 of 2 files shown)


V3RON (commented):
Oh, I'm not sure what the reason is, but Xcode uses a 'bundle.js' wrapper over the React Native CLI instead of calling the CLI directly. That's why bundle mode detection is failing. We should check whether the executable is 'scripts/bundle.js' from 'react-native'.

Jheysoon:
Updated the PR. Thanks
Also tested the change in local and it works

PR #99, opened 2025-09-08
fix(metro): prevent hanging in EAS builds

V3RON:
Currently, if the Redux DevTools plugin is used, the build process hangs when using EAS, since another CLI is involved in orchestrating the build. This quick fix is intended to prevent that and give me time to develop a more resilient solution.


V3RON:
Closing in favour of #100.

Issue #126, opened 2025-10-16
[redux-devtools-plugin] Not listing physical device vs emulator + Hermes asyncIterator problems


dprevost-LMI:
### Describe the bug

We integrated Rozenite into our application, and it generally works well beside the Redux plugin, which shows nothing.
We usually run on physical devices, and after generating an RN 82 test project and comparing the emulator vs. physical devices, we found that the redux plugin works well on the emulator but not on the physical device (At least for Android; iOS has not been tested yet)

Working Redux rozenite plugin screenshot (Android emulator)
<img width="1431" height="986" alt="Image" src="https://github.com/user-attachments/assets/5a5f486d-f863-41f0-bd2e-bcef98b849fd" />

Non-working case on physical device (screenshot using Vysor)
<img width="1644" height="907" alt="Image" src="https://github.com/user-attachments/assets/7374e29a-6016-4cad-bb4f-ce2ff223ab52" />

Both are on the same project/code, which is a new RN 82 project generated by the react-native-community/cli with a basic integration of Roxzenite + redux/toolkit. See https://github.com/dprevost-LMI/rn82/tree/rozenite-and-redux.

### System Info

```shell
info Fetching system and libraries information...
System:
  OS: macOS 15.6.1
  CPU: (12) arm64 Apple M2 Pro
  Memory: 96.61 MB / 32.00 GB
  Shell:
    version: "5.9"
    path: /bin/zsh
Binaries:
  Node:
    version: 20.19.4
    path: /Users/dprevost/.nvm/versions/node/v20.19.4/bin/node
  Yarn:
    version: 1.22.22
    path: /Users/dprevost/.nvm/versions/node/v20.19.4/bin/yarn
  npm:
    version: 10.8.2
    path: /Users/dprevost/.nvm/versions/node/v20.19.4/bin/npm
  Watchman:
    version: 2025.08.18.00
    path: /opt/homebrew/bin/watchman
Managers:
  CocoaPods:
    version: 1.16.2
    path: /Users/dprevost/.rbenv/shims/pod
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 24.2
      - iOS 18.2
      - macOS 15.2
      - tvOS 18.2
      - visionOS 2.2
      - watchOS 11.2
  Android SDK: Not Found
IDEs:
  Android Studio: 2025.1 AI-251.26094.121.2513.14007798
  Xcode:
    version: 16.2/16C5032a
    path: /usr/bin/xcodebuild
[...]

PR #127, opened 2025-10-18
fix(redux-devtools-plugin): asyncIterator not defined

dprevost-LMI:
As found out in this [issue](https://github.com/callstackincubator/rozenite/issues/126#issuecomment-3416711486), adding documentation on the Redux DevTool plugin for a physical device + Hermes when we cannot select the device in the inspector's selector, either because:
1. Data on port 8765 is not proxied because of a missing adb reverse
2. Potential Hermes bug with `asyncIterator` in Redux DevTools
3. Move the init of `asyncIterator` before its usage in the import of `redux-devtools/remote`

### DIFF
--- a/.nx/version-plans/version-plan-1761752669919.md
+++ b/.nx/version-plans/version-plan-1761752669919.md
@@ -1,0 +1,5 @@
+---
+'@rozenite/redux-devtools-plugin': prerelease
+---
+
+Fixed a race condition that occurred in certain scenarios by inlining the require call to the redux-devtools/remote package.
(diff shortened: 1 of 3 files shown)


V3RON:
Have you tried the solution I posted earlier in the related issue? It should resolve the problem cleanly, without any weird imports in the app code.

```
export const rozeniteDevToolsEnhancer = (): StoreEnhancer => {
  const { devToolsEnhancer } = require("@redux-devtools/remote");
  return devToolsEnhancer({
    name: getDeviceId(),
    hostname: getHostname(),
    port: REDUX_DEVTOOLS_PORT,
    secure: false,
    realtime: true,
  });
};
```

dprevost-LMI:
Oh yes, [this post](https://github.com/callstackincubator/rozenite/issues/126#issuecomment-3431825843), I skipped it and forgot to take a mental note and focus on the post after instead 😒 

And yes, it works nicely, updating PR...

image

V3RON:
Thank you for the contribution! 👏

Issue #82, opened 2025-08-23
Network requests fired before plugin initialization are not captured


manish2612:
### Describe the bug

When using @rozenite/network-activity-plugin in a React Native 0.78.2 project, the plugin successfully displays network requests in the React Native DevTools once it is initialized. However, any network requests that are fired before the plugin is initialized (e.g. during the initial render of the root component or from global singletons that execute at module load) are not captured in the Network Activity panel.

This means that the very first screen’s API calls are consistently missing from the logs, even though later requests show up as expected.

Expected Behavior

All network requests — including those fired during the initial render phase or before React mounts the root component — should be intercepted and displayed in the Network Activity panel.

Actual Behavior

Requests made before plugin initialization are silently missed.

Only requests after initialization are visible in DevTools.

### System Info

```shell
React Native: 0.78.2

@rozenite/network-activity-plugin: latest

DevTools: new RN DevTools (not Flipper)

Platform: iOS
```

### Rozenite Version

1.0.0-alpha.11

### Reproduction

NA

### Steps to reproduce

Add @rozenite/network-activity-plugin via Metro config (withRozenite).

Initialize the plugin using useNetworkActivityDevTools() at the very top of the root component (before rendering anything else).

Trigger a network request (e.g. via fetch or axios) immediately in the first screen’s useEffect.

Observe that this initial request does not appear in the Network Activity panel.

Any subsequent requests (triggered after initialization) do appear correctly.

V3RON:
Hey @manish2612 👋 
There is a noticeable delay between the app starting and the DevTools session being established. By design, the plugin stays dormant as long as no connection is present, in order to avoid putting unnecessary pressure on the CPU and memory. The downside of this approach is essentially what you described.

The built-in React Profiler mitigates this issue with its "restart and profile" functionality, which instructs React Native to restart and start profiling regardless of whether a debugging session is active. Technically, the Network Activity plugin could adopt the same approach, but the problem is that there is no built-in API for persisting data between full reloads. I also don't want to force users to install additional native packages just to use Rozenite.

I do have an idea for making this behavior completely optional, and I'll investigate it this week. However, with React Universe Conf approaching, my capacity is limited, so I can't promise it will be ready by Sunday.

dprevost-LMI:
If no one has started anything, I could —we need this, too!

dprevost-LMI:
Here a potential PR solving this issue: https://github.com/callstackincubator/rozenite/pull/143

[...]

PR #143, opened 2025-11-15
feat(network-activity-plugin): show http network requests on boot

dprevost-LMI:
## Description
Fixes https://github.com/callstackincubator/rozenite/issues/82. 
Using a queuing mechanism, we can queue boot clients' messages destined to the DevTool and display them once the inspector panel is connected and is ready.

- I added a generic interface for all the inspectors
- I use an event listener to switch from a queue to the client, and kept it generic to be reused in other plugins
- I renamed the former network-inspector to http and created a new network-inspector managing the three types (http, ws,sse)
- I tried not needing withOnBootNetworkActivityRecording, but I was unable to initialize everything sooner.
- Align some HTTP code with how SSE and WS were done, like the events in a separate type and file.

TODO
- [x] revert code added temporarily to have a working playground
- [x] Add official docs
- [x] Review if we keep laststart and in progress event....
- [x] Review last TODOs
- [x] Test without the `withOnBootNetworkActivityRecording`
- [x] Review implementation following code review change request

## Related Issue
See https://github.com/callstackincubator/rozenite/issues/82

## Context
We heavily rely on the WebSocket and HTTP requests on boot for our enterprise app, so we totally need this feature, so I took a go for it.

## Testing

Tested as below, plus using unit test coverage yet to open in another PR, see preview in [this PR](https://github.com/dprevost-LMI/rozenite/pull/1)

https://github.com/user-attachments/assets/d900744b-46d5-40f3-951e-fbd988e0d602

---

> [!NOTE]
> Adds boot-time HTTP request capture via a queued client and `withOnBootNetworkActivityRecording`, includes progress/timeout support and UI updates, plus docs and playground examples.
> 
> - **Network Activity Plugin (core)**:
>   - **Boot-time capture**: New `withOnBootNetworkActivityRecording` and queued client (`queued-client-wrapper`) to buffer events before DevTools connects; `network-inspector` updated to flush queue on enable.
>   - **Refactor**: Extract per-request logic to `request-tracker` (tracking, overrides, response body helpers).
>   - **Events**: Add `request-progress` (with loaded/total), handle `timeout`; wire through store (`model`, `derived`, `store`) and client (`shared/client`).
>   - **UI**: `RequestList` shows in-progress percent; `Toolbar` swaps record/stop icon logic and adds tooltip; default recording enabled.
> - **Playground**:
>   - Extract real API to `apps/.../utils/network-activity/api.ts`; add "Large File" download flow using `api.getLargeFile` and UI tab; `main.tsx` enables on-boot recording and triggers sample requests.
> - **Docs**:
>   - Update README and website to document boot-time recording usage.
> - **Tooling**:
>   - Add repo scripts; upgrade Prettier to v3; broaden `.prettierignore` to `**/dist` and `**/coverage`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 020a939d9c0bb6f506742d03e474e7ae814a65d5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


apps/playground/src/app/App.tsx
```diff
@@ -12,7 +12,6 @@ import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
 import { useRef } from 'react';
 import { SafeAreaProvider } from 'react-native-safe-area-context';
 import { Provider } from 'react-redux';
-import { mmkvStorages } from './mmkv-storages';
```
dprevost-LMI: To revert, removed since playground was not booting brecause of sqllite problem

apps/playground/src/app/store.ts
```diff
@@ -1,13 +1,10 @@
 import { configureStore } from '@reduxjs/toolkit';
 import counterReducer from './store/counterSlice';
-import { rozeniteDevToolsEnhancer } from '@rozenite/redux-devtools-plugin';
 
 export const store = configureStore({
   reducer: {
     counter: counterReducer,
   },
-  enhancers: (getDefaultEnhancers) =>
-    getDefaultEnhancers().concat(rozeniteDevToolsEnhancer()),
```
dprevost-LMI: To revert, removed since playground was not booting brecause of sqllite problem

package.json
dprevost-LMI: Waiting on merge of #130

V3RON:
Unfortunately, it's not going to work in some cases. There is a slight delay between the client becoming available and the other side being ready to accept messages. In other words, useRozeniteDevToolsClient may return a client instance and allow you to send messages, but the other side may not be listening yet.

I need to refactor the bridge to include some sort of handshake mechanism so both sides can confirm that they’re ready to handle communication correctly.

dprevost-LMI:
> Unfortunately, it's not going to work in some cases. There is a slight delay between the client becoming available and the other side being ready to accept messages. In other words, useRozeniteDevToolsClient may return a client instance and allow you to send messages, but the other side may not be listening yet.
> 
> I need to refactor the bridge to include some sort of handshake mechanism so both sides can confirm that they’re ready to handle communication correctly.

I performed a few iterations and flushed the queue when I received the `network-enable` signal.
For sure, having a "more intelligent" client would be better across the board

dprevost-LMI:
I have reached a point where this solution, although not entirely finalized, is the one I suggest.
After that, a fully persisted solution would be the next step!

apps/playground/src/main.tsx
```diff
@@ -3,4 +3,17 @@ import App from './app/App';
 
 import 'react-native-get-random-values';
 
+import { withOnBootNetworkActivityRecording } from '@rozenite/network-activity-plugin/react-native';
+import { api } from './app/utils/network-activity/api';
+
+withOnBootNetworkActivityRecording();
```
dprevost-LMI: I wonder if there is a way to auto-inject that 🤔

dprevost-LMI: After some research, I do not think so!

packages/network-activity-plugin/src/react-native/http/network-inspector.ts
```diff
@@ -190,151 +57,20 @@ export type NetworkInspector = {
...
-    XHRInterceptor.setSendCallback(handleRequestSend);
-    XHRInterceptor.setOverrideCallback(handleRequestOverride);
-    XHRInterceptor.enableInterception();
+    if (queuedClient.isBootInterceptionEnabled()) {
+      // Boot interception is already active, just switch mode and flush queue
+      queuedClient.enableClient();
+      queuedClient.flushQueue();
+    } else {
+      setupXHRInterceptor(queuedClient);
+    }
```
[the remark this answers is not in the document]
dprevost-LMI: Not sure if this was a real problem, but I reviewed the code to be more robust and make sure this never happens.

V3RON (changes requested):
This pull request includes not only recording network traffic on boot, but also changing the way how 'long requests' are displayed (status with progress). Please clean it up, so only a single feature remains.

For the implementation, instead of combining traffic inspection and queueing logic, I suggest the following approach:
a) refactor inspectors to share the same API (emit events)
b) provide a special type of an inspector accumulating events and flushing them on demand
c) detect the presence of 'queued inspector' and flush events when plugin client is ready (in practice, when DevTools UI sends some kind of 'hello' message)

```typescript
type Inspector<TEventMap> = {
  enable: () => void;
  disable: () => void;
  on<TEventType extends keyof TEventMap(type: TEventType, callback: (event: TEventMap[TEventType]) => void);
  on(type: '*', callback: (event: TEventMap[keyof TEventMap]) => void);
};

type NetworkInspector = Inspector<NetworkEventMap>;

type QueuedInspector = Inspector & {
  flush: () => void;
};

const networkInspector = createRozeniteNetworkInspector({
  recordOnBoot: true,
});
const websocketInspector = createRozeniteWebSocketInespector();

const useNetworkActivityDevTools = ({ inspectors }) => {
  const pluginClient = useRozeniteDevToolsClient({ pluginId: '...' });
  useEffect(() => {
    inspectors.forEach((inspector) => {
      inspector.on('*', (event) => pluginClient.send(event.type, event));
      
      if ('flush' in inspector) {
        inspector.flush();
      }
    });
  }, [inspectors, pluginClient]);
}

const App = () => {
  useNetworkActivityDevTools({
    inspectors: [networkInspector, websocketInspector]
  });
}
```
Note: this is pseudo-code, not production-ready 😄 

In this approach, the inspector only handles inspection and passing data to the listeners. Queuing is not its concern.

dprevost-LMI:
I'll proceed to the change as you requested!

Just so you know: One thing I was thinking about with this queued client is that if we can make it generic enough, it could be exposed as a generic interface, allowing any package to use it directly and be boot-time compliant.

dprevost-LMI:
> Please clean it up, so only a single feature remains.

Done, see https://github.com/callstackincubator/rozenite/pull/147 and also https://github.com/callstackincubator/rozenite/issues/150

> For the implementation, instead of combining traffic inspection and queueing logic, I suggest the following approach:
> a) refactor inspectors to share the same API (emit events)
> b) provide a special type of an inspector accumulating events and flushing them on demand
> c) detect the presence of 'queued inspector' and flush events when plugin client is ready (in practice, when DevTools UI sends some kind of 'hello' message)

a) I think I had partially done it with `request-tracker.ts`, I'll look again to see how I can do it with network-inspector 🤔.
b) That sounds similar to my queue-client-wrapper, will look to move that into the inspector instead 🤔.
c) That was done in `inspector.enable()` called when the DevTool is ready, and sends the `network-enable`

I might reach out to you tomorrow on Discord since it seems blurry in my mind right now!

dprevost-LMI:
@V3RON, seeing that rn 83 will provide a network panel by default ([see change logs](https://github.com/facebook/react-native/blob/main/CHANGELOG.md#v0830-rc3) and screenshots), then I'm not sure that iterating on this solution is a good use of time. So, what I see:

1. Finishing the solution I provided without significant changes, since, anyway, with RN 83, this plugin will be "dead"
2. Close the PR and wait on RN 83

What do you think?

Screenshot 2025-11-29 at 6 50 15 AM

After discussion, even though they provide a network, but probably not WebSockets or SSE, then it is worth continuing.

dprevost-LMI:
👋 Hey, I'm back to it.
I would like to iterate on the architecture point by point to move forward; otherwise, I'm not sure how to move forward and align with the vision described above.

**Brainstorming in progress**

The first point is the queue used by the client/inspector
So the way I envision this queue is tightly coupled with the client and would be a generic way that is possibly even hidden from the initial hook `useNetworkActivityDevTools`, and therefore not needing the interface below, since I do not wish for the inspector to have to differentiate between on boot and not on boot or the regular client vs the queue client.

```ts
type QueuedInspector = Inspector & {
  flush: () => void;
};
```
In other words, my thought (even though the code does not fully reflect it) is 
1. By default, we have a queue client that supports on-boot communication.
2. Make it generic enough so we can actually reuse it for other plugins, and so it should not be tied to the inspector.
3. To achieve this, it makes sense to actually queue the message that the client would normally send to the Dev Tools, but accumulate in a queue.
4. Then once the hook starts, it actually creates and configures the plugin's real client and requests to be used, at that moment we know soon we will be able to communicate with the devTools and send the message and flush the queue.
5. However, passing the client is not enough; we also need a "ready" communication from the Dev Tool.
6. Having a generic "ready" built-in ack message in the plugin client/Dev tool would be the best, actually, to leverage this queue for any other rozenite plug-in, but this is starting to be out of scope of this PR.
7. In this case, it will be on `'network-enable'`. I could use this as a configuration for the onBoot client, where it would be the generic "ready" message, but configurable. 🤔 

So, to elaborate technically:

App.tsx
```ts
const networkInspector = createRozeniteNetworkInspector({
  recordOnBoot: true, // Probably false by default to ensure previous behaviour and no regression?
  onBootBuffer: 1000,
});
 ...other inspector

const App = () => {
  useNetworkActivityDevTools({
    inspectors: [networkInspector, websocketInspector]
  });
}
```

useNetworkActivityDevTools.ts
```ts
const useNetworkActivityDevTools = ({ inspectors }) => {
  const pluginClient = useRozeniteDevToolsClient({ pluginId: '...' });
  useEffect(() => {
    inspectors.forEach((inspector) => {
      inspector.setPluginClient(client);
  }, [inspectors, pluginClient]);
}
```

OnBootClient.ts
```ts
export class OnBootQueuingClient {
private messageQueue: QueuedMessage[] = [];
private onFlush: (pluginClient) => {
   for event in messageQueue {
    pluginClient.send(event.type, event)
  }
}

/**
* High-level view:
* 1. Queue message
* 2. Once the real client is ready, stop queuing and flush
* 3. If possible, be transparent from the hooks and the client initialization/setup
**/

}
```

createRozeniteNetworkInspector.ts
```ts
const createRozeniteNetworkInspector = ( {  
  recordOnBoot: false,
  onBootBuffer: 100
}
) => {
   const onBootClient = new OnBootQueuingClient({recordOnBoot, onBootBuffer)
   const networkInspector = getNetworkInspector(onBootClient)
   return networkInspector
}
```
network-inspector.ts
``` ts
export const getNetworkInspector = (
  onBootClient,
): {
// Start XMLInterception
// use onBootClient as the client listening to the request events
return new NetworkInspector(onBootClient);
}

class NetworkInspector {
  private client = onBootClient();
  export const setPluginClient = (
    pluginClient: NetworkActivityDevToolsClient,
  ): {
      const enableSubscription = pluginClient.onMessage('network-enable', () => {
          onBootClient.onFlush(pluginClient)
      });
     // How to switch from onBootClient to pluginClient
  }
}
```

I'll continue brainstorming the solution as I write this post.
One thing I have realized now is that I needed the client Wrapper to use a singleton in the `XHRInterceptor`, else stopping `XHRInterceptor` to switch clients can also break ongoing requests. Hence, we need a client orchestrator that will not change for the interceptor but still switch from the queue version to the plugin client.

Maybe you have already covered that IN the design you mentioned earlier, and I'm yet to see it 😅 

I'll start with what I mentioned above and reiterate the post during the day!


Update:
I vibe code a new solution that seems way better, more aligned, and recording HTTP, SSE and WS on boot.
It is still unfinished and maybe not fully aligned with the changes request, but it is working at first glance

image

Work will continue in the upcoming hours or days, let's see!

dprevost-LMI:
@V3RON the PR should be ready now:
1. I added the suggested interface
2. I use an event-listener, maybe not entirely as you requested, but kept generic to be reused in other plugins
3. I rename the former network-inspector to http and create a new network-inspector managing the three types (http, ws,sse)
4. I tried not needing `withOnBootNetworkActivityRecording` but I was unable to initialize everything sooner.
5. Align some HTTP code with how SSE and WS were done, like the events in a separate type and file.

V3RON (approved):

PR #222, opened 2026-04-09
feat: Rozenite for Agents SDK

manapard:
## Description

This PR expands Rozenite for Agents from a CLI-only workflow into a reusable SDK and shared agent platform, while also shipping a new SQLite plugin and improving web support/docs.

### Agent SDK and agent platform refactor
- adds the first public `@rozenite/agent-sdk` package for programmatic Rozenite agent workflows
- exposes:
  - `createAgentClient()`
  - `client.withSession(...)`
  - `client.openSession()`
  - `client.attachSession()`
  - `session.domains.*` and `session.tools.*`
  - low-level HTTP transport via `@rozenite/agent-sdk/transport`
- refactors the existing `rozenite agent` CLI to use the SDK internally
- moves reusable agent logic out of the CLI into shared packages and middleware-backed HTTP routes
- keeps support for:
  - static + runtime domain discovery
  - domain/tool resolution
  - optional auto-pagination for paged tools

### Typed agent tool contracts
- adds shared typed tool contract/descriptor helpers in `@rozenite/agent-shared`
- updates `@rozenite/agent-bridge` so tool handler args/results can be inferred from typed contracts
- adds plugin `./sdk` entrypoints for typed tool descriptors, including `@rozenite/storage-plugin/sdk`
- updates the build pipeline to emit clean per-target SDK declaration bundles


packages/agent-sdk/src/types.ts
```diff
@@ -0,0 +1,97 @@
...
+    listTools: (input: {
+      sessionId: string;
+      domain: string;
+    }) => Promise<AgentDomainTool[]>;
+    getToolSchema: (input: {
+      sessionId: string;
+      domain: string;
+      tool: string;
+    }) => Promise<AgentToolSchema>;
+    callTool: (input: {
```
V3RON: We can make it generic `<TArgs, TResult>`, so it's easier to provide types later on.

packages/agent-sdk/package.json
```diff
@@ -0,0 +1,56 @@
...
+  ],
+  "homepage": "https://github.com/callstackincubator/rozenite#readme",
+  "bugs": {
+    "url": "https://github.com/callstackincubator/rozenite/issues"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/callstackincubator/rozenite.git"
+  },
+  "author": "Szymon Chmal <szymon.chmal@callstack.com>",
```
V3RON: I'm not an author 😛 Feel free to put yourself here!

packages/agent-sdk/src/types.ts
V3RON: I wonder if there's a way to safely expose tools from plugins without bringing in any JavaScript or third-party packages. We'd probably need a separate entry point, like `@rozenite/storage-plugin/sdk`

V3RON:
```
1. High: packages/agent-sdk/src/pagination.ts:88-100,128-129
   maxItems is not enforced on the first page.
   The helper seeds merged.items with all of initial.items before any truncation logic runs, and only trims subsequent pages. If the first response already exceeds maxItems, the SDK returns too many items.
   Validation:
   packages/middleware/src/agent/runtime/pagination/paginate.ts:81-89 and packages/network-activity-plugin/src/react-native/agent/state.ts:199-205 show paged tools may legitimately return up to the requested page limit on page 1, so the SDK must enforce its own client-side cap consistently.
   Test gap:
   packages/agent-sdk/src/__tests__/pagination.test.ts:50-77 only covers truncation after fetching page 2, not an oversized first page.
   
2. Medium: packages/agent-sdk/src/pagination.ts:117-124
   hasMore can become incorrect when maxItems truncates the merged result.
   After truncating, the code unconditionally sets merged.page.hasMore = true. If the last fetched page actually had hasMore: false and no nextCursor, the SDK returns pagination metadata that implies another page exists when it does not.
   Validation:
   The shared pagination contract in packages/middleware/src/agent/runtime/pagination/types.ts:9-20 and implementation in packages/middleware/src/agent/runtime/pagination/paginate.ts:69-89 show hasMore and nextCursor are coupled server facts. Fabricating hasMore: true client-side can produce an impossible state.
   Test gap:
   No test covers truncation on a terminal page.
```

packages/cli/skills/rozenite-agent-sdk/agents/openai.yaml
V3RON, line 1: What's that? 👀

packages/cli/skills/rozenite-agent-sdk/SKILL.md
```diff
@@ -0,0 +1,60 @@
...
+1. Write or open a small Node ESM script.
+2. Import `createAgentClient()` from `@rozenite/agent-sdk`.
+3. Start with `client.withSession(...)` unless the session must stay open across separate steps.
+4. If multiple devices may be connected, call `client.targets.list()` and pass `deviceId`.
+5. Use `session.domains.list()` to confirm what is registered on the live target.
+6. If needed, inspect a domain with `session.tools.list({ domain })` or inspect one tool with `session.tools.getSchema({ domain, tool })`.
+7. Call the needed tool with `session.tools.call(...)`, either by name or with a typed plugin descriptor.
+8. If paged results should be merged, pass `autoPaginate`.
+9. If the session must stay open, switch to `client.openSession()` and stop it in `finally`.
+10. If reconnecting to an already-existing session, use `client.attachSession(sessionId)`.
```
V3RON: I think agent should use 'rozenite agent' to get the list of potential targets. My thinking is that this step may fail because of reasons and it may be more efficient to just stop there instead of letting agent write code and then try to debug it.

packages/cli/skills/rozenite-agent-sdk/SKILL.md
```diff
@@ -0,0 +1,60 @@
...
+## Choosing Calls
+
+- Use a call by name when you know the domain and tool name, or when you discovered them at runtime.
+- Use a typed plugin call when a plugin publishes descriptors under `./sdk` and type inference will save effort or reduce mistakes.
+
+## Managing Sessions
+
+- Use `withSession(...)` for almost all short scripts and one-off tasks.
+- Use `openSession()` only when the session must stay open across multiple independent steps or function boundaries.
+- Use `attachSession(sessionId)` only when reconnecting to an existing session instead of creating a new one.
```
V3RON: Generally, for the sake of token efficiency, let's only describe 'withSession', which should internally create session when needed or reuse existing one.

packages/cli/skills/rozenite-agent-sdk/SKILL.md
V3RON, line 1: Let's treat tool descriptors as go-to way for agents to use tools. This will shrink SKILL.md significantly.

V3RON (approved):

PR #226, opened 2026-04-12
feat(sqlite-plugin): add adapter-helpers subpath export

Mohamed-kassim:
### Summary

We’re integrating Rozenite SQLite inspector with non-Expo SQLite drivers (e.g. react-native-nitro-sqlite). The plugin already has shared SQL normalization / statement classification / bridge decode / error formatting helpers, but they were not available on a stable public subpath.

This PR exposes those helpers via a dedicated subpath:

- `@rozenite/sqlite-plugin/adapter-helpers`

This keeps helper code out of the main package entry and avoids leaking it into production builds that only use the core plugin APIs.

### Implementation notes

- Added `src/react-native/adapter-helpers.ts` to re-export existing shared helpers from `src/shared/sql` and `src/shared/bridge-values`.
- Added `./adapter-helpers` to `packages/sqlite-plugin/package.json` exports.
- Added a second RN build entry in `vite.config.ts` so dist emits:
  - `dist/react-native/adapter-helpers.js`
  - `dist/react-native/adapter-helpers.cjs`
  - `dist/react-native/adapter-helpers.d.ts`

### Question for maintainers

We also have a small adapter that wires react-native-nitro-sqlite into `createSqliteAdapter`, using these same helpers. Are you open to follow-up PRs that add additional first-party adapters (e.g. Nitro)?

---

Thanks for reviewing.

### DIFF
--- a/packages/sqlite-plugin/package.json
+++ b/packages/sqlite-plugin/package.json
@@ -76,6 +76,11 @@
       "import": "./dist/react-native/index.js",
       "require": "./dist/react-native/index.cjs"
     },
+    "./adapter-helpers": {
+      "types": "./dist/react-native/adapter-helpers.d.ts",
+      "import": "./dist/react-native/adapter-helpers.js",
+      "require": "./dist/react-native/adapter-helpers.cjs"
+    },
     "./package.json": "./package.json"
   },
   "publishConfig": {
--- a/packages/sqlite-plugin/src/react-native/adapter-helpers.ts
+++ b/packages/sqlite-plugin/src/react-native/adapter-helpers.ts
@@ -1,0 +1,11 @@
+/**
+ * Shared SQL + bridge helpers for custom adapters (e.g. non-Expo SQLite drivers).
+ */
+export type { SqlStatementSegment } from '../shared/sql';
+export {
+  classifySqlStatement,
+  normalizeSingleStatementSql,
+  splitSqlStatements,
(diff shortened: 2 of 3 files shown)


V3RON (commented):
I think we should create a separate entry point for these helpers, since they are currently being pulled into production app builds (they are not gated by the if statement below).

For example, we could define a new entry point like `@rozenite/sqlite-plugin/adapter-helpers` and export them from there. This would eliminate any risk of leakage.

Mohamed-kassim:
> I think we should create a separate entry point for these helpers, since they are currently being pulled into production app builds (they are not gated by the if statement below).
> 
> For example, we could define a new entry point like `@rozenite/sqlite-plugin/adapter-helpers` and export them from there. This would eliminate any risk of leakage.

Sorry for the delay, I missed your reply.
Addressed! I moved the helper exports out of the main entry into a dedicated subpath. The main entry is now focused on the existing plugin API.

V3RON:
This approach doesn't work quite as expected. Vite crashes when trying to run vite-require-plugin because it generates multiple outputs where only one is expected. Even after fixing that, some extra files still end up in dist/react-native, like the JS bundle for the DevTools UI.

To keep things simple for now, I decided to go with the same approach we already use for other plugin exports and just provide mocks in production. It's not perfect, but it should work well enough until I get a chance to revisit the Vite plugin setup.

See #259

PR #261, opened 2026-05-11
feat(storage-plugin): add storage-level JSON import/export

### LINKED ISSUE #247: Add storage-level JSON import/export to the storage plugin
Storage plugin users need a simple way to move a storage snapshot in and out of the inspector during development. This is useful for reproducing bugs, sharing app state between teammates, seeding known test scenarios, restoring a previously captured state after reinstalling or clearing data, and speeding up manual QA workflows without writing one-off app code.

Today the plugin supports inspecting and mutating entries one by one, but it does not provide a storage-level import/export workflow. That makes larger state setup and state sharing unnecessarily slow, especially for storages that hold many related keys.

## Observed Findings

- The plugin already models storage data as typed entries, with support for `string`, `number`, `boolean`, and `buffer` values depending on storage capabilities.
- The panel works at the selected-storage level and already keeps full snapshots of the selected storage in memory.
- The runtime/plugin bridge already supports setting and deleting entries, which makes import as an upsert feasible without introducing a separate storage mutation model.
- Different adapters support different value types. For example, MMKV can expose `buffer` entries, while AsyncStorage and Expo SecureStore are string-only.
- Current validation already respects per-storage supported types in the UI/runtime. Import must keep the same guarantee for file-based workflows.
- The current feature set is entry-oriented: users can add, edit, delete, search, and inspect entries, but they cannot export a storage snapshot to a file or import one back.

## Suggested Behavior

Add storage-level import/export for the currently selected storage only.

Export:
- Allow users to export the currently selected storage to a JSON file.
- The exported file should use a structured, versioned snapshot format rather than a backend-native dump.
- The format should preserve typed entries so it remains portable across supported storage adapters where types are compatible.
- Storage metadata may be included for context and diagnostics, but export should still represent a single selected storage only.

Import:
- Allow users to import a JSON snapshot file into the currently selected storage.
- Import should behave as an upsert:
  - If a key from the file already exists, overwrite it.
  - If a key from the file does not exist, create it.
  - Keys not present in the file should remain untouched.
- Import must validate the entire file before writing anything.
- If any entry in the imported file has a type unsupported by the currently selected storage, reject the import and surface a clear error to the user.
- The selected storage is the import target. File metadata should not be used to redirect the import elsewhere.

Recommended JSON shape:

```json
{
  "version": 1,
  "plugin": "@rozenite/storage-plugin",
  "createdAt": "2026-05-07T12:34:56.000Z",
  "storage": {
    "adapterId": "mmkv",
    "storageId": "user",
    "adapterName": "MMKV",
    "storageName": "user",
    "capabilities": {
      "supportedTypes": ["string", "number", "boolean", "buffer"]
    }
  },
  "entries": [
    { "key": "token", "type": "string", "value": "abc" },
    { "key": "launchCount", "type": "number", "value": 3 },
    { "key": "seenOnboarding", "type": "boolean", "value": true },
    { "key": "blob", "type": "buffer", "value": [1, 2, 3] }
  ]
}
```

Implementation requirements:
- Define and document a stable, versioned JSON snapshot schema for one storage.
- Validate file structure and entry shapes before import starts.
- Validate that every imported entry type is supported by the selected storage.
- Abort the whole import if validation fails.
- Apply a successful import as entry upserts.
- Update the panel state after import so the imported result is visible immediately.
- Expose the feature in the storage plugin UI in a way that is clearly scoped to the currently selected storage.
- Document the import/export behavior, schema, and compatibility expectations in the storage plugin docs.

Out of scope for this issue:
- Global export/import across multiple storages in one file.
- Replace/clear-storage semantics during import.
- Adapter-specific native dump formats.

## Resolution Summary

The storage plugin should let users export the currently selected storage to a structured JSON snapshot and import such a snapshot back into the currently selected storage as an upsert, while rejecting incompatible files when any imported entry type is not supported by that storage.

burczu:
Closes #247.

## Summary

- Adds **Import** and **Export** buttons to the storage panel toolbar, scoped to the currently selected storage.
- Defines a stable, versioned (`version: 1`) JSON snapshot format with `plugin`, `createdAt`, source `storage` metadata, and typed `entries[]`.
- Hand-written `parseSnapshot` with path-precise errors (e.g. `entries[3].value: Expected finite number…`). No new runtime deps.
- Import is an **upsert** with **validate-before-write**: any unsupported type aborts before any writes. The preview shows new / overwrite / skipped (blacklisted) counts and warns when the file's source storage differs from the target.
- Runtime delivers a new `import-entries` event; emits a `set-entry` echo per successful write so the table updates row by row, then a single `import-result` for completion. Stops at the first failed write and reports the offending key.
- Extends the snapshot wire format with an optional serialized `blacklist` (RegExp `source`/`flags`) so the panel can compute skip counts against the runtime's filter.
- Docs: new **Import / Export** section in the storage plugin docs covering behavior, schema, cross-adapter compatibility, and versioning. Pointer added to the package README.
- Changeset: minor bump for `@rozenite/storage-plugin`.

## Test plan

Automated:

- [x] Snapshot parser unit tests — happy paths, every rejection branch, path precision (`packages/storage-plugin/src/shared/__tests__/snapshot.test.ts`).
- [x] `computePreview` pure-function tests — bucket priority, blacklist skip, unsupported types, metadata mismatch.
- [x] Runtime bulk handler tests — per-write echo, stop-on-first-error, unknown target, empty entries (`packages/storage-plugin/src/react-native/__tests__/import.test.ts`).
- [x] `pnpm --filter @rozenite/storage-plugin test` — 48/48 passing.
- [x] `pnpm --filter @rozenite/storage-plugin typecheck` — clean.
- [x] `pnpm --filter @rozenite/storage-plugin lint` — 0 errors (1 pre-existing unrelated warning).
- [x] `pnpm --filter @rozenite/docs build` — website builds; new section renders.

Manual (playground, MMKV / AsyncStorage / SecureStore via `apps/playground`):

- [x] Export downloads a valid JSON file named `rozenite-storage-<adapter>-<storage>-YYYYMMDD-HHmmss.json`.
- [x] Importing an exported file (unchanged) is a no-op preview that applies cleanly.
- [x] Importing a file with `version: 2` is rejected with a path-pointed error.
- [x] Importing into an adapter that doesn't support a present type (e.g. `buffer` into AsyncStorage) disables **Apply** and lists offenders.
- [x] Importing from a different storage source triggers the metadata-mismatch warning banner.
- [x] Importing an empty `entries: []` file applies as a no-op.
- [x] Importing valid edits updates the table row by row as writes land.

## Out of scope

- Global cross-storage export.
- `replace-all` / clear-storage semantics during import.
- Adapter-specific native dump formats.
- Surfacing storages that use `shouldFilterKey` (function form) — only the `blacklist` RegExp form is exposed to the UI today.

### DIFF
--- a/packages/storage-plugin/src/react-native/import.ts
+++ b/packages/storage-plugin/src/react-native/import.ts
@@ -1,0 +1,67 @@
+import type {
+  StorageImportEntriesEvent,
+  StorageImportResultEvent,
+  StorageSetEntryEvent,
+} from '../shared/messaging';
+import type { StorageView } from './storage-view';
+
+export type ImportEmittedEvent =
+  | StorageSetEntryEvent
+  | StorageImportResultEvent;
+
+export const handleImportEntries = async (
+  views: StorageView[],
+  event: StorageImportEntriesEvent,
+  emit: (out: ImportEmittedEvent) => void,
+): Promise<void> => {
+  const view = views.find(
+    (candidate) =>
+      candidate.target.adapterId === event.target.adapterId &&
+      candidate.target.storageId === event.target.storageId,
+  );
+
+  if (!view) {
+    emit({
+      type: 'import-result',
+      target: event.target,
+      ok: false,
+      written: 0,
+      total: event.entries.length,
+      error: 'Target storage not found',
+    });
+    return;
+  }
+
+  for (let i = 0; i < event.entries.length; i++) {
+    const entry = event.entries[i];
+
+    try {
+      await view.set(entry);
+    } catch (error) {
+      emit({
+        type: 'import-result',
+        target: event.target,
+        ok: false,
+        written: i,
+        total: event.entries.length,
+        failedKey: entry.key,
+        error: error instanceof Error ? error.message : String(error),
+      });
+      return;
+    }
+
+    emit({
+      type: 'set-entry',
+      target: event.target,
+      entry,
+    });
+  }
+
+  emit({
+    type: 'import-result',
+    target: event.target,
+    ok: true,
+    written: event.entries.length,
+    total: event.entries.length,
+  });
+};
--- a/.changeset/quiet-snakes-export.md
+++ b/.changeset/quiet-snakes-export.md
@@ -1,0 +1,7 @@
+---
+'@rozenite/storage-plugin': minor
+---
+
+Add storage-level JSON import/export to the storage plugin.
+
+You can now export the currently selected storage to a versioned JSON snapshot and import a snapshot back as an upsert. Import validates the file before writing and rejects entries whose types are not supported by the target storage. See the storage plugin docs for the schema and behavior.
--- a/packages/storage-plugin/README.md
+++ b/packages/storage-plugin/README.md
@@ -103,3 +103,4 @@ When you use [Rozenite for Web](https://rozenite.dev/docs/rozenite-for-web) in d
 - Type support is enforced in UI, runtime and Agent tools.
 - Storages without subscriptions automatically use internal polling updates.
 - Per-storage blacklists are supported through adapter configuration.
+- Storage snapshots can be exported to and imported from a versioned JSON file from the panel. See the [storage plugin docs](https://www.rozenite.dev/docs/official-plugins/storage) for the schema and behavior.
--- a/packages/storage-plugin/src/react-native/__tests__/import.test.ts
+++ b/packages/storage-plugin/src/react-native/__tests__/import.test.ts
@@ -1,0 +1,182 @@
+import { describe, expect, it, vi } from 'vitest';
+import { handleImportEntries, type ImportEmittedEvent } from '../import';
+import type { StorageView } from '../storage-view';
+import { getStorageViewId } from '../../shared/types';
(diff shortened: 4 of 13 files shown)


packages/storage-plugin/src/react-native/import.ts
V3RON, line 1: Is there a specific reason for handling imports like this instead of sending a batch of set-entry commands from the DevTools UI side?

burczu: `client.send` is fire-and-forget — no per-message ack. With N × `set-entry`, the UI knows it dispatched N messages but never learns whether the writes landed or where they failed, which makes the issue's "stop on first error" and the modal's "Imported 47/50, failed at `userToken`" outcome impossible to honor.

`import-entries` + `import-result` puts the loop, the stop-on-first-error rule, and the final `{ ok, written, total, failedKey, error }` payload in one place. Per-write `set-entry` echoes still flow back, so the table updates row by row through the existing snapshot path.

Wire cost is two small event types, both unit-tested. Happy to switch if you'd rather accept silent partial failures.

PR #266, opened 2026-05-12
feat(network-activity-plugin): add advanced request filters

### LINKED ISSUE #241: Add advanced request filters to Network Activity
Network Activity users need better ways to narrow down busy request logs when debugging real React Native apps. The current panel is good for small sessions, but once an app loads multiple screens, prefetches data, talks to several backends, and mixes HTTP with WebSocket or SSE traffic, a single free-text filter is not enough to isolate the request a developer cares about.

## Observed Findings

The Network Activity plugin already captures a rich request list for HTTP, WebSocket, SSE, and Nitro traffic. The panel supports a free-text filter and a request-type filter for HTTP / WS / SSE, but it does not support Chrome DevTools-style property filters such as method, status code, domain, MIME type, size, source, failure state, or in-flight state.

## Suggested Behavior

The Network Activity panel should support advanced filters that let users quickly isolate relevant traffic without scanning the full table manually.

From a user perspective, filtering should cover at least:
- HTTP method
- Status code or status range
- Domain / host
- Request source such as Built-in or Nitro
- Request type / transport
- Failed requests only
- In-flight requests only
- Overridden requests only
- Content type / MIME type
- Size and duration thresholds

The feature should feel like a natural extension of the existing filter bar, not a separate power-user tool hidden elsewhere. Users should be able to combine multiple filters in a single session and see the request list update immediately.

The filter UI should be understandable for casual users while still scaling to large logs. That could mean structured chips, dropdown-driven filters, or a property-query syntax, as long as the resulting experience makes it easy to answer questions like:
- show me all failed POST requests
- show me only Nitro traffic
- show me slow requests to a specific backend
- show me only JSON responses from one host

Filters should work consistently across the request list and should remain compatible with the existing basic text search experience.

## Resolution Summary

Network Activity should let users filter traffic using practical request properties, so developers can narrow large React Native network logs down to the exact subset they need during everyday debugging.

draggie:
## Description

Adds advanced request filtering to Network Activity so busy request logs can be narrowed by practical request properties instead of only free-text search and transport type.

The filter bar now supports method, request source, status code/range, domain, MIME type, failed-only, in-flight-only, overridden-only, size thresholds, and duration thresholds. Existing text search and type filters continue to work and can be combined with the new advanced filters.

## Related Issue

Closes #241

## Context

The request list already had most of the data needed to support richer filtering, but some values were only available in detailed entries. This change extends the processed request summary with response content type and SSE source, then normalizes table rows once and applies all filters against those row values.

Status filters support exact codes, ranges such as , classes such as , and comparisons such as . Size and duration thresholds use raw byte and millisecond values so filtering remains predictable.

## Testing
Verified manually in the playground that the advanced filters work and the existing clear-filters button resets the combined filter state.
image

### DIFF
--- a/.changeset/network-activity-advanced-filters.md
+++ b/.changeset/network-activity-advanced-filters.md
@@ -1,0 +1,5 @@
+---
+"@rozenite/network-activity-plugin": minor
+---
+
+Add advanced request filters to Network Activity.
--- a/packages/network-activity-plugin/src/ui/components/FilterBar.tsx
+++ b/packages/network-activity-plugin/src/ui/components/FilterBar.tsx
@@ -1,14 +1,20 @@
+import { useState } from 'react';
+import {
+  autoUpdate,
+  flip,
+  FloatingPortal,
+  offset,
+  shift,
+  size,
+  useClick,
+  useDismiss,
+  useFloating,
+  useInteractions,
+  useRole,
+} from '@floating-ui/react';
 import { Input } from './Input';
 import { Button } from './Button';
-import { X, Filter, ChevronDown } from 'lucide-react';
-import {
-  DropdownMenu,
-  DropdownMenuTrigger,
-  DropdownMenuContent,
-  DropdownMenuCheckboxItem,
-  DropdownMenuLabel,
-  DropdownMenuSeparator,
-} from './DropdownMenu';
+import { X, Filter, ChevronDown, Check } from 'lucide-react';
 import type { HttpMethod, NetworkEventSource } from '../../shared/client';
 
 export type RequestTypeFilter = 'http' | 'websocket' | 'sse';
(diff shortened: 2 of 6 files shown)


V3RON:
I believe there is no need for a separate 'type' filter. We should move it to your 'Advanced' dropdown and probably rename it to 'Filters' or reducing the trigger to an icon only. WDYT?

draggie:
@V3RON 
image
I would keep it as button with title

V3RON:
There is one more thing:
We pre-select all options for 'Request Type'. Should we default to 'nothing selected' and treat is as 'all selected'? I think that's the logic now for 'Request Source' and others. WDYT?

draggie:
Final version has request type filter inverted: 
image

V3RON:
@draggie 
We have to address two things before we release it:
1. There is no max-height set for this dropdown, so it tends to overflow the viewport
2. We should not close the dropdown on every click - we should wait for backdrop click. Right now it's annoying to have to click on 'Filters' again just to set the second filter.

V3RON (approved):

Issue #244, opened 2026-05-07
Improve response viewing in Network Activity


V3RON:
Network Activity users need richer response inspection because modern React Native apps return more than plain JSON and raw text. Debugging often involves images, HTML error pages, formatted JSON, XML payloads, large text responses, or binary placeholders where the current experience does not provide enough context to understand what actually came back from the server.

## Observed Findings

The Network Activity plugin already shows response bodies for supported HTTP requests and formats JSON and plain text reasonably well. However, the current experience is still limited compared to Chrome DevTools. Binary responses are reduced to a placeholder, richer previews are missing, and the response view does not fully support the range of response formats developers regularly inspect while working on React Native apps.

## Suggested Behavior

The Network Activity panel should provide a more capable response-viewing experience for captured requests.

From a user perspective, the response area should better support:
- formatted JSON that is easy to inspect and navigate
- readable text and XML responses
- HTML responses that are easier to understand when the server returns an error page
- image responses with an actual preview when feasible
- clearer metadata for binary or non-previewable responses
- large responses in a way that remains usable instead of feeling like a raw text dump

The improved response view should help users answer questions like:
- did the server actually return the image I expected
- is this HTML error response a backend failure or an unexpected redirect page
- what exactly is inside this deeply nested JSON payload
- is this response binary, compressed, empty, or simply not previewable in the current viewer

The UI should make the distinction between preview and raw response understandable when both are useful. The overall experience should feel closer to a full inspection tool and less like a fallback text viewer.

## Resolution Summary

Network Activity should provide a richer response viewer that handles common real-world response formats more clearly, so React Native developers can understand server responses without leaving the panel.

burczu:
## Plan

After looking at the current response viewer and the six concerns the issue raises (richer JSON, readable text/XML, HTML preview, image preview, binary metadata, large responses), I'm proposing five PRs.

**Merge order:** PR1 lays the foundation and must merge first. PR2/3/4 add format support on top and can land in any order after PR1. PR5 is a self-contained performance improvement and can ship at any point — including before PR1.

### PRs

**PR1 — Image preview, Preview/Raw toggle, and the format-renderer architecture** *(foundational, must merge first)*

The response viewer learns to display images inline (PNG, JPEG, GIF, WebP, SVG) via a real `<img>` element — SVG renders straight from its text source, the rest ship as base64-encoded bytes with a 5MB safety cap (above which the panel shows a clear "Response too large for preview" message instead of choking).

[...]

PR #271, opened 2026-05-15
feat(network-activity-plugin): image preview + renderer registry

burczu:
First PR of #244 — \"Improve response viewing in Network Activity\". Lays the architectural foundation the next PRs in the series build on, and ships image response preview as the first user-visible capability.

## Summary

- **Renderer registry architecture**. `ResponseTab` becomes a thin dispatcher over a small array of self-contained format renderers. Each renderer declares `matches(contentType, body)`, the views it offers (`['preview']` / `['raw']` / both), its default view, whether it supports the existing Override flow, and how to render each view. Future PRs in this series (rendered HTML, formatted XML, JSON Raw, binary hex, virtualization) each append one or two entries without touching the wrapper.
- **Image response preview**. Inline `` for `image/png`, `image/jpeg`, `image/gif`, `image/webp` rendered from base64 data URLs. `image/svg+xml` carved onto the text path so the Raw tab shows source XML rather than base64. Bitmap images get a metadata card (Content-Type + decoded size) on the Raw tab. **Preview / Raw sub-tab toggle** is adaptive — hidden when a format only has one meaningful view — and remembers the user's last choice across requests in a panel session.
- **Additive wire-format extension.** The `response-body` bridge event's `body` widens from `string | null` to a discriminated union: `string | { kind: 'binary'; base64 } | { kind: 'binary-too-large'; size } | null`. Existing consumers that only handle the string variant continue to work unchanged. The 5 MB in-capture size cap short-circuits to the \`binary-too-large\` variant; oversized image responses show *\"Response too large for preview (X MB)\"* without crossing the bridge.
- **Capture lockstep.** XHR (`http/http-utils.ts`) is updated to emit binary union variants for image content-types. The nitro-fetch capture (`nitro-fetch/nitro-network-inspector.ts`) type widens to the same shape; its runtime emits string-only until a native-side change in \`react-native-nitro-fetch\` adds binary support, tracked separately.
- **Tests.** 55 vitest passes across dispatcher unit tests, capture unit tests (jsdom), and React Testing Library tests for the image / SVG / binary-too-large renderers and the `ViewToggle` component. RTL is set up at the package level for the first time so PR2-5 inherit the pattern.
- **Playground.** A new \"Images\" tab on the Network Test screen fires GETs at PNG / JPEG / WebP / SVG / a large JPEG / a deliberately oversized JPEG (Hubble Pillars of Creation, ~45 MB) so the new behavior can be exercised end-to-end.

## Test plan

- [ ] \`pnpm --filter @rozenite/network-activity-plugin test\` — 55 tests pass
- [ ] \`pnpm --filter @rozenite/network-activity-plugin typecheck\` — clean
- [ ] \`pnpm --filter @rozenite/network-activity-plugin lint\` — clean
- [ ] \`pnpm --filter @rozenite/network-activity-plugin build\` — succeeds (prod-bundle sanity)
- [ ] Playground (\`pnpm start:playground:ios\` or \`:android\`) → Network Test → Images tab:
  - [ ] PNG / JPEG / WebP — Preview shows inline image, Raw shows metadata card, \`[Preview | Raw]\` toggle visible
  - [ ] SVG — Preview renders the vector, Raw shows source XML (not base64)
  - [ ] Large JPEG — Preview renders cleanly, under-cap behavior
  - [ ] Huge JPEG (cap test) — \"Response too large for preview\", toggle hidden
- [ ] Sticky preference: toggle once to Raw on one image, click a different image in the request list, stays on Raw
- [ ] Non-image regression check: HTTP Test → users / posts / todos responses still render as JsonTree; text responses render as CodeBlock; toggle hidden in both cases
- [ ] Override flow on JSON/text still works; Override button **absent** on image responses (and on the binary-too-large placeholder)

## Followups

- **Nitro binary capture.** Native-side change in \`react-native-nitro-fetch\` to surface binary response bytes alongside text. Until then, image responses captured via nitro fall through to the text path.
- **PR2** of #244: non-image binary hex viewer + Download button + metadata card upgrade.
- **PR3-5** of #244: HTML rendering, XML tree + JSON Raw, large-text virtualization. Each appends to the renderer array introduced here.


V3RON (approved):
👏

PR #275, opened 2026-05-15
feat(network-activity-plugin): HTML response renderer with sandboxed iframe

burczu:
Part of #244 — "Improve response viewing in Network Activity." Adds an HTML renderer that previews captured `text/html` responses inside a sandboxed iframe.

## Summary

- **`htmlRenderer`** (`src/ui/response-renderers/html.tsx`) — matches `text/html` (case-insensitive via `normalizeContentType`, charset-tolerant). Registered between `jsonRenderer` and `textFallbackRenderer`. Does **not** claim `application/xhtml+xml` (left for a future XML renderer).
- **Preview** renders the captured response inside `<iframe sandbox="" srcDoc=...>`. `sandbox=""` (empty value = all restrictions on) is the primary defense — scripts cannot execute and the iframe is treated as a unique origin.
- **CSP** as defense-in-depth — a `<meta http-equiv="Content-Security-Policy" content="default-src 'none'; style-src 'unsafe-inline'; img-src data:;">` is prepended to `srcDoc`. This blocks subresource fetches that `sandbox` does not block on its own (external images, stylesheets, fonts). Without CSP, previewing arbitrary HTML would fire unsolicited requests from the developer's browser to whatever URLs the captured response references — surprising privacy side-effect for a DevTools panel.
- **Raw** view renders the HTML source via the existing `CodeBlock` — same treatment as text-fallback and SVG Raw.
- **Status banner** for error responses (`status >= 400`) renders inside Preview only, above the iframe ("Server returned 404 NOT FOUND" with a red border-left). In Raw view the user already reads source directly; the banner becomes noise. The banner is HTML-renderer-specific because only HTML can visually misdirect — a styled 5xx "Whoops" landing page looks intentional until you see status. JSON/XML/text show source either way.
- **`RenderCtx` grows two optional fields** — `status` and `statusText` — plumbed from `selectedRequest.response` in `ResponseTab`. Other renderers ignore them.
- **`supportsOverride: true`** — HTML is a string; the existing override editor works on it.
- **Playground** — a new "HTML" main-tab on the Network Test screen with three demo buttons (pre-verified URLs):
  - `Classic HTML` — `https://httpbin.org/html` (200, ~3.7 KB Moby Dick excerpt)
  - `Minimal HTML` — `https://example.com/` (200, inline styles — verifies `style-src 'unsafe-inline'`)
  - `404 HTML` — `https://httpbin.org/nonexistent-rozenite-test-route` (404, triggers the banner)

## Test plan

- [ ] `pnpm --filter @rozenite/network-activity-plugin test` — all tests pass (16 new in `html.test.tsx` + dispatch additions).
- [ ] `pnpm --filter @rozenite/network-activity-plugin typecheck` — clean.
- [ ] `pnpm --filter @rozenite/network-activity-plugin lint` — clean.
- [ ] `npx prettier --check` on changed files — clean.
- [ ] Playground (`pnpm --filter @rozenite/playground ios` or `:android`) → Network Test → HTML tab:
  - [ ] **Classic HTML** — Preview shows the rendered Moby Dick page. Toggle to Raw → CodeBlock with HTML source. Toggle back → iframe re-renders.
  - [ ] **Minimal HTML** — example.com renders with its inline-styled heading visible.
  - [ ] **404 HTML** — red banner above the iframe reads "Server returned 404 NOT FOUND". Banner disappears in Raw view, returns when toggling back to Preview.
- [ ] **Sticky toggle** — toggle Raw on one HTML response, click a different captured request, the new response opens in Raw too. Toggle to Preview on one → next response also opens in Preview.
- [ ] **Override button** visible on HTML responses (Pencil icon, top right of the Section).
- [ ] **Sandbox sanity** — no scripts run inside the iframe. None of the three demo URLs include `<script>`, so this is implicit; for a hard check use a local HTML response that contains `<script>document.body.innerHTML="XSS"</script>` and confirm the body shows the unmodified source rendered via the parser without script execution.

## New test surface

- `response-renderers/__tests__/html.test.tsx` (16) — renderer attributes; preview iframe attribute coverage (`sandbox=""`, CSP meta prefix on `srcDoc`, body preserved verbatim); status banner thresholds (200/399 absent; 400/404/500 present; Raw view never; missing `status` absent); XSS-attempt regression group asserting the iframe attributes hold even when the body contains `<script>` (guards against future "simplifications" of the sandbox config).
- `response-renderers/__tests__/dispatch.test.ts` (+5 assertions) — `text/html` routes to `html` (flipped from `text-fallback`); `text/html; charset=utf-8` and `text/HTML` (uppercase) route to `html` (prove normalization); `application/xhtml+xml` still routes to `text-fallback` (not claimed); empty-string `text/html` body routes to `html`. Plus a registry-order invariant (`html` precedes `text-fallback`).

## Followups (deliberately not in this PR)

- Syntax-highlighted HTML source view (cross-cutting; pick a highlighter for HTML/JSON/XML/etc. in a separate PR).
- Configurable iframe size / resize handle / size persistence.
- `application/xhtml+xml` handling — left for the future XML renderer.

## Notes

- The CSP meta tag injection via `srcDoc` is "best effort": browsers honor pre-doctype `<meta http-equiv>` in srcdoc reliably in practice, but the formal spec leaves wiggle room. `sandbox=""` is the load-bearing defense; CSP narrows the side-effect surface.
- jsdom does not execute iframe `srcDoc`, so "the script did not run" cannot be unit-tested. The XSS regression test asserts the iframe **attributes** that prevent execution remain unchanged. Real behavioral verification is manual playground work.


V3RON:
Looks great!

I have only one comment: maybe we should get rid of this banner as HTTP status is already shown in the header, just after the URL?
Screenshot 2026-05-19 at 07 31 12

burczu:
Sure, I'll remove it and resolve conflicts shortly

PR #308, opened 2026-06-15
feat(network-activity-plugin): support Expo fetch interception

### LINKED ISSUE #295: Expo 56 expo/fetch network activity support
### Description

While `expo/fetch` has been available for a bit, in Expo 56 it is by default installed as the global `fetch` handler.

https://docs.expo.dev/versions/v56.0.0/sdk/expo/#expofetch-api

### Suggested solution

_No response_

### Additional context

_No response_

V3RON:
## Description

Add Expo SDK 54+ fetch traffic to Network Activity. Explicit `expo/fetch` calls are captured across the supported range, and Expo SDK 56's default global fetch alias is captured through the same interceptor. Expo requests include request and response metadata, available bodies, failures, cancellation state, and an Expo source label in the existing Network Activity UI.

## Related Issue

Closes #295.

## Context

The public `expo/fetch` entry exposes a getter-only re-export, so the interceptor resolves and patches Expo's writable private fetch implementation instead. The optional resolver is isolated in its own build chunk so React Native applications without Expo retain the existing XHR behavior.

SDK 56 responses use clone-based body capture. SDK 54–55 responses, where cloning is unsupported, capture common bodies when application code consumes `text()` or `arrayBuffer()`; unconsumed and directly streamed bodies remain unavailable. The wrapper preserves original response and rejection identity, restores private and global fetch aliases safely, and leaves XHR and Nitro interception independent.

## Testing

- `pnpm --filter @rozenite/network-activity-plugin test` — 197 tests passed
- `pnpm --filter @rozenite/network-activity-plugin typecheck`
- `pnpm --filter @rozenite/network-activity-plugin lint`
- `pnpm --filter @rozenite/network-activity-plugin build`
- `pnpm release:plan`
- `git diff --check`
- Inspected the build output: the private Expo require is isolated in `dist/react-native/chunks/get-expo-fetch-module.{js,cjs}`, and the main React Native entry contains no Expo import or require.
- Manual Expo SDK 54/56 device smoke tests and a bare React Native Metro smoke test remain to be performed.

### DIFF
--- a/.changeset/expo-fetch-network-activity.md
+++ b/.changeset/expo-fetch-network-activity.md
@@ -1,0 +1,6 @@
+---
+'@rozenite/network-activity-plugin': minor
+---
+
+Show Expo SDK 54+ `expo/fetch` activity in Network Activity, including Expo's
+SDK 56+ default global fetch alias.
--- a/apps/playground/src/app/screens/NetworkTestScreen.tsx
+++ b/apps/playground/src/app/screens/NetworkTestScreen.tsx
@@ -25,6 +25,10 @@ import {
 import { RootStackParamList } from '../navigation/types';
 import { api, User, Post, Todo } from '../utils/network-activity/api';
 import {
+  expoFetchApi,
+  type ExpoFetchDemoResult,
+} from '../utils/network-activity/expo';
+import {
   nitroApi,
   type NitroDemoResult,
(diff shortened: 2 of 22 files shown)


mzaien:
@V3RON any updates on this ?

PR #312, opened 2026-06-26
feat(file-system-plugin): add open in file manager button in details panel

ManojSankolSR:
- Added a host-side POST /open-in-file-manager endpoint to the middleware server to open local folders on macOS, Windows, and Linux.
- Updated the DevTools details panel to display an "Open in File Manager" button next to "Export" for both selected files and directories.
- Kept directory selection active when navigating inside to allow host reveals.
- Implemented a graceful WebSocket-based simulator/device fallback when running on physical/non-local environments.

### DIFF
--- a/.changeset/add-open-in-file-manager-button.md
+++ b/.changeset/add-open-in-file-manager-button.md
@@ -1,0 +1,6 @@
+---
+"@rozenite/file-system-plugin": minor
+"@rozenite/middleware": minor
+---
+
+Add "Open in File Manager" button in the Details Panel to reveal files/folders in the host desktop file manager (Finder/Explorer), with a device-side fallback.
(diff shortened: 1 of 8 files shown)


V3RON:
Hey! I looked through the PR and the functionality looks great. One question: how is this expected to work on Android, where emulator files live inside the AVD image and aren’t directly accessible from the host, unlike iOS Simulator files?

ManojSankolSR:
@V3RON   the Android implementation handles this entirely on the device side. Instead of pinging the middleware to open the host's file manager, the plugin uses React Native's Linking.openURL() with an ACTION_VIEW intent (using either a content:// or file:// URI). So when you click the button for an Android app, it actually launches the native file manager (or whatever default app is set) inside the emulator or physical device itself

I understand that opening it on the device isn't exactly the expected approach, but due to the current limitations, this was the most feasible implementation.

Issue #320, opened 2026-07-28
Expose one shared cursor-based pagination primitive for plugin authors


V3RON:
## Summary

Agent tool pagination is implemented four different ways across the codebase, the shared primitive that should prevent that is duplicated in two private copies and exported to nobody, and the most common of the four conventions — offset/limit over a live collection — **silently skips rows**.

Proposal: settle on **one** pagination model (cursor/keyset), move the existing primitive into `@rozenite/agent-shared`, export it for plugin authors, and add the one adapter that makes the simple cases a one-liner.

### Four conventions

| where | shape |
| --- | --- |
| `mmkv`, `storage`, `file-system`, `redux-devtools` plugins | `offset` / `limit`, hand-rolled |
| `network-activity` plugin | `cursor` / `limit`, hand-rolled |
| `tanstack-query` plugin | `cursor` + `page{hasMore,nextCursor}` |
| built-in `console` domain | `paginateSource` with a validated opaque cursor |

An agent that learns how `console.getMessages` pages learns nothing transferable about `mmkv.list-entries`.

### The primitive is duplicated and private

`paginateSource`, `PaginatedSource`, `encodeCursor`, `hashFilters` exist in **two** copies:

- `packages/cli/src/commands/agent/runtime/pagination/`
- `packages/middleware/src/agent/runtime/pagination/`

`diff` shows the divergence is currently **formatting only** (different prettier settings), so consolidation is mechanical today. It will not stay that way.

Neither copy is exported from `agent-shared`, `agent-sdk`, or `agent-bridge`, so a plugin author has no way to reach it. That is why six plugins hand-rolled their own.

## Motivation: offset pagination over a live app is incorrect

`packages/mmkv-plugin/src/react-native/useMMKVAgentTools.ts:87-90`:

```js
const allKeys = view.getAllKeys();
const safeOffset = Math.max(0, Math.floor(offset));
const keys = allKeys.slice(safeOffset, safeOffset + safeLimit);
```

Same pattern in `useStorageAgentTools.ts:108-110` and `useFileSystemAgentTools.ts:98-107`. **No sort anywhere.** Two independent defects:

1. **Order is undefined.** `getAllKeys()` ordering is an implementation detail of the underlying store. Nothing guarantees two calls agree, so "page 2" is not a well-defined concept.

[...]

PR #322, opened 2026-07-28
feat(cli): emit reusable columnar JSON for paginated output

### LINKED ISSUE #318: Emit columnar JSON for row-shaped agent output to cut token usage
## Summary

Agent-facing CLI output currently emits row-keyed JSON, which repeats every field name on every row. For the row-shaped payloads that dominate agent usage (tree nodes, request lists, storage keys, domains, tools) this is the single largest avoidable cost in the output path.

Proposal: encode row-shaped results columnar — a `cols` header plus positional `rows` — instead of an array of objects. Measured up to **65% fewer bytes** on narrow-valued tables.

This is independent of transport (CLI, SDK, or a future MCP server) and can ship on its own.

## Current output

`printOutput` in `packages/cli/src/commands/agent/output.ts` always emits JSON (the `-j/--json` flag is currently a documented no-op). Listings are wrapped in `{items, page}` by `paginateRows` in `output-shaping.ts`.

```json
{"items":[{"nodeId":1,"label":"App","displayName":"App","elementType":"function","childCount":1},{"nodeId":4,"label":"NavigationRoot","displayName":"NavigationRoot","elementType":"function","childCount":2,"parentId":1}],"page":{"limit":20,"hasMore":false}}
```

## Proposed output

```json
{"cols":["nodeId","label","displayName","elementType","childCount","parentId"],"rows":[[1,"App","App","function",1,null],[4,"NavigationRoot","NavigationRoot","function",2,1]]}
```

Spaced out for readability:

```json
{
  "cols": ["nodeId", "label", "displayName", "elementType", "childCount", "parentId"],
  "rows": [
    [ 1, "App",            "App",            "function", 1, null],
    [ 4, "NavigationRoot", "NavigationRoot", "function", 2,    1]
  ]
}
```

Nested values remain first-class — a cell can hold an object or array, so `--verbose` selections that include `context` or `argsPreview` need no special handling:

```json
{
  "cols": ["seq", "timestamp", "level", "source", "context", "text"],
  "rows": [
    [1, 1730000000000, "info", "console", null, "App mounted"],
    [2, 1730000000100, "error", "exception",
      {"stack": "App.tsx:42", "componentStack": "ProductList"},
      "Warning: Each child in a list should have a unique \"key\" prop.\n    in ProductList (at App.tsx:42)"]
  ]
}
```

Payloads that are not row-shaped (`react.getComponent`, `network.getRequestDetails`, acks) stay exactly as they are. Compact JSON is already at the size floor for nested data — measured alternatives were worse: YAML-style indentation +3%, pretty-printed JSON +75%.

## Measurements

Minified byte counts, comparing `{items:[...]}` against `{cols,rows}` for the same records:

| rows | tree nodes (short values) | console messages (one long text field) |
| --- | --- | --- |
| 1 | **−7%** | **−4%** |
| 2 | 28% | 9% |
| 5 | 51% | 18% |
| 10 | 59% | 20% |
| 25 | 63% | 22% |
| 50 | 64% | 23% |
| 100 | 65% | 23% |

Two conclusions:

**Single-row results get worse.** The `cols` header costs more than one row of field names saves. Columnar should only kick in at **2 or more rows**; single-row results stay row-keyed.

**The saving is very uneven, and depends on the ratio of field-name bytes to value bytes.** Tree-node rows are short integers and identifiers, so names dominate and the saving approaches 65%. Console rows carry one long `text` field that dwarfs its keys, so the saving plateaus around 23%. Expect the headline number on `getTree` / `searchNodes` / `listRequests` / `list-entries`, and roughly a quarter of it on console-heavy work.

Related: for console specifically, the larger lever is projection, not encoding. `--fields` / `--verbose` exist for the domain and tool *listings* (`parseFields` in `output-shaping.ts`) but not for tool call results, which are the large payloads. Trimming default fields on `getMessages` would save more than the encoding change does. Worth doing alongside, but tracked as its own concern.

## Also in scope: drop the page envelope

`paginateRows` emits `{"page":{"limit":20,"hasMore":false}}` even when there is nothing more to fetch — pure overhead on every terminal page.

Proposal: omit `page` when `hasMore` is false, and when it is true replace the bare cursor with the next command. Agents copy commands more reliably than base64 tokens, and the cursor is only tamper-resistant against a caller who has no reason to tamper.

```json
"next": "rozenite agent call console.getMessages --cursor eyJ2IjoxLCJzZXEiOjJ9"
```

## Determinism requirement

Whatever the selection rule ends up being, **the output shape must be a function of `(tool, selected fields, row count)` and never of the observed values.** If it depended on whether any record happened to carry a nested field, the same command with the same flags would return different shapes depending on what the app logged — which is far more confusing to a consumer than either shape on its own.

## Suggested scope

- [ ] Add a columnar encoder to `output-shaping.ts`, applied at ≥2 rows
- [ ] Leave non-row payloads untouched (no envelope, no re-encoding)
- [ ] Omit `page` when `hasMore` is false; emit `next` as a runnable command when true
- [ ] Repurpose the currently no-op `--json` flag to mean "row-keyed / expanded shape" as an escape hatch
- [ ] Update `agent-output-format.test.ts` and `agent-output-shaping.test.ts`
- [ ] Update `skills/rozenite-agent/SKILL.md` and `domains/*.md` examples to the new shape
- [ ] Document which shape is a stable contract and which may change

## Compatibility

Nothing first-party parses CLI stdout — `@rozenite/agent-sdk` talks HTTP directly via `transport.ts`, so it is unaffected.

However, this **does** change the default stdout shape, which is a breaking change for anyone parsing it today. Two options worth discussing:

1. Ship columnar as the new default and give `--json` its meaning back as the expanded/legacy shape (single release, breaking)
2. Introduce the columnar shape behind an opt-in flag first, then flip the default in a later major

## Notes

Filed without a label — this is a performance/output-format change and the repo has no matching label.

V3RON:
## Summary

- add a public paginated tool declaration to `@rozenite/agent-shared`, re-exported from `@rozenite/agent-bridge`
- propagate pagination field metadata through tool registration, multi-device aggregation, SDK discovery, descriptors, and schema inspection
- replace the CLI's hard-coded domain/tool allowlist with metadata-driven shaping that works for built-in and third-party plugins
- migrate built-in React, console, and network tools plus the Network Activity and TanStack Query plugins
- document the external plugin contract and include the release changeset

## Behavior

Declared paginated tools return the normal `{items, page}` contract. Their registered metadata defines the stable allowed/default field order; the CLI never infers columns from observed values.

- two or more rows emit `cols` / `rows`
- zero or one row remains row-keyed under `items`
- absent optional values become `null`
- tool-owned metadata is preserved
- terminal pagination metadata is omitted
- continued pages receive a shell-safe runnable `next` command
- undeclared, malformed, and non-row results remain unchanged
- `--json` remains a deprecated no-op with no legacy escape hatch

## Why

Agent CLI row payloads repeated field names for every record, increasing byte and token usage. The original implementation solved this with a CLI-private allowlist, which excluded third-party plugins and duplicated knowledge of tool schemas. The corrected design makes pagination presentation an explicit tool contract owned by each producer.

Issue #320 remains responsible for the separate shared cursor/source pagination engine and migrations away from offset pagination.

## Validation

- 108 focused pagination, CLI, bridge, SDK, middleware, and plugin tests pass
- affected package typechecks pass
- affected package lint checks pass
- `@rozenite/agent-bridge` production build passes
- Prettier and `git diff --check` pass
- changeset release plan validates
- the broader CLI run retains five unrelated existing config-wrapper expectation failures caused by the generated `enabled` option

Closes #318


V3RON:
Merged `main` into this branch and pushed fixes for a code review pass. Fast-forward from `1fc93326`; no history rewritten.

## Review findings addressed

| Commit | Fixes |
|---|---|
| `29faa712` | merge `main` (6 commits) — one trivial conflict in `agent-local-domains.test.ts`, an import this branch's new test needs |
| `f7c2cd1f` | every paginated call made two full `getSessionTools` fetches (`getSchema` then `call`, both through the uncached `resolveDomainContext`) — replaced with a single `session.tools.resolve()`. Non-object `--args` was silently replaced with `{}`; now throws. Malformed third-party `pagination` metadata failed the whole call; now degrades to unshaped output, matching the documented contract |
| `cd1f5c2c` | `next` echoed the user-typed domain token, so a deprecated legacy slug (`at-rozenite__mmkv-plugin`, kept "for one release cycle" per #328) could be baked into a generated command — now emits the canonical `domain.id`. `next` also emitted a bare `rozenite ...` while `SKILL.md` instructs `npx rozenite` |
| `98e836e8` | `projectRows` filled `null` in the row-keyed branch, making 0/1-row payloads *larger* than before — which is the whole reason that branch exists (#318 measures columnar at −7% for one row). Also: `page.reset` was handled by the CLI but no producer ever emitted it, so a stale cursor after an app relaunch returned `{"items":[]}`, indistinguishable from "no more rows" |
| `987dd84a` | no first-party tool declared `defaultFields`, so `--verbose` was a no-op and `console.getMessages` always emitted `argsPreview` + `context` — the opposite of #318's "projection is the bigger lever". Built-in `pagination.fields` were unchecked strings; now coupled to their row types via `keyof`, so drift is a build error |
| `12c802c7` | moved the encoder, projection, and page envelope into `@rozenite/agent-shared` per the request on #318, leaving the CLI a thin renderer. `formatAgentCommand`/`shellEscape` stay CLI-side per carve-out 1, and the shared shaper takes a frontend-neutral affordance instead of assuming a shell command |

## Notes

Two findings from the review were split out rather than folded in here:

- **#331** — `packages/cli` has no `test` script, so `turbo run ... test --affected` never ran this package's tests, including the ~600 lines this PR adds. It also explains the 5 `config-wrapper.test.ts` failures that had been sitting in the repo. Worth landing first so this PR's tests actually execute in CI.
- **#332** — `autoPaginate.maxItems` drops rows and returns a cursor that skips them. Pre-existing in `agent-sdk`, but this PR is what turns that cursor into a runnable `next` command. Once it lands, this branch should consume the new `page.truncated` signal and suppress `next` when it is set.

Carve-out 2 (token-based rather than row-based page cap) is **not** addressed here, deliberately — see my note on #320.

Also worth a reviewer's eye: `98e836e8` changes the network cursor `scope` from `network:requests:${generation}` to `network:requests` with `generation` as its own field, so a stale cursor now returns `page.reset: true` instead of a generic "Invalid cursor" error. And `cd1f5c2c` changes the tools-listing cursor scope from the typed token to the canonical id, without which a re-run of this PR's own emitted command would fail scope validation.

## Validation

- `npx turbo run typecheck lint test --force` across the whole repo: **100 successful, 100 total**
- `packages/cli` suite directly (turbo does not run it, see #331): 10 files, 86 tests, all passing
- New `@rozenite/agent-shared` test suite: 16 tests — that package previously had none
- Verified the field-drift protection by introducing a typo: `TS2820: Type '"nodeIdXX"' is not assignable to type 'keyof ReactTreeNode'`
- `prettier --check` fails on 1654 files, identically on `main`; the two files this branch had regressed are fixed in `4bbb7bcc`

One caveat on the earlier `tsc --noEmit -p tsconfig.json` checks in this branch's history: `packages/cli/tsconfig.json` and `packages/middleware/tsconfig.json` are solution files with `"include": []`, so that command typechecks nothing. Everything above was re-verified through turbo, which builds workspace dependencies first.

PR #332, opened 2026-07-30
fix(agent-sdk): stop maxItems from dropping rows and reporting a stale cursor

### LINKED ISSUE #329: Remove SDK auto-pagination; preserve plugin-owned pages and cursors
## Summary

`callToolWithOptionalPagination` in `packages/agent-sdk/src/pagination.ts` trims the collected items down to `maxItems`, but copies the producer's page envelope through unchanged. The resulting `page.nextCursor` points past the rows that were just discarded, so those rows become unreachable — no error, no gap marker.

There are two independent loss paths.

**1. Initial page** (`pagination.ts:97-101`)

```js
const merged: PagedResponse = {
  ...initial,
  items: getTrimmedItems(initial.items, maxItems),  // trims to maxItems
  page: { ...initial.page },                        // keeps the full page's nextCursor
};
```

With `limit: 20, maxItems: 5`, the tool returns 20 rows, the caller receives 5, and `merged.page.nextCursor` still points past row 19. Rows 5–19 are silently unreachable. The `while` loop never runs, because its `merged.items.length < maxItems` guard is already false.

**2. Mid-loop** (`pagination.ts:132-140`)

```js
const remaining = maxItems === undefined
  ? next.items.length
  : Math.max(0, maxItems - merged.items.length);

merged.items.push(...next.items.slice(0, remaining));
merged.page = { ...next.page };   // same defect
```

## Steps to reproduce

Against any cursor-paginated agent tool (e.g. `console.getMessages`) over a 20-row dataset:

1. Call with `autoPaginate: { pagesLimit: 2, maxItems: 5 }` and `limit: 20`.
2. Observe 5 items returned, plus a `page.nextCursor`.
3. Follow that cursor.
4. The union of both responses is missing rows 5–19.

Mid-loop variant: `limit: 5, maxItems: 7, pagesLimit: 2` over the same dataset returns rows 0–6, and the returned cursor resumes at row 10 — rows 7, 8, 9 are lost.

## Why it matters now

The cursor used to be an opaque token that nothing acted on automatically. #318 makes the CLI render it into output as a runnable `npx rozenite agent ... --cursor <token>` command, so a lossy cursor becomes a copy-pasteable instruction that skips data.

## Note

The SDK cannot fix this by synthesizing a cursor — cursor encoding is producer-owned and opaque, so there is no way to express "row 5 of this page". The fix is to stop over-fetching: request only the rows still within budget, so a well-behaved producer's own cursor lands exactly on the truncation boundary. A producer that clamps or ignores `limit` still needs an explicit "resume position unknown" signal rather than a misleading cursor.

This is independent of #320, which covers the producer-side pagination model.

V3RON:
## Description

`callToolWithOptionalPagination` trimmed collected items down to `maxItems` but passed the producer's page envelope through unchanged, so the returned `page.nextCursor` pointed past the rows it had just discarded. Those rows became unreachable with no error and no gap marker.

Two independent loss paths, both fixed:

- **Initial page** — with `limit: 20, maxItems: 5`, the tool returned 20 rows, the caller received 5, and the cursor still pointed past row 19. Rows 5–19 vanished, and the result looked complete.
- **Mid-loop** — with `limit: 5, maxItems: 7`, rows 7–9 were fetched, discarded, and skipped by the propagated cursor.

The fix is to stop over-fetching: every request, including the first, now asks for `min(callerLimit, maxItems - collectedSoFar)`, so a well-behaved producer's own cursor lands exactly on the truncation boundary. No cursors are synthesized.

For producers that clamp or ignore `limit` (`console.getMessages` documents "Default 50, max 200"), truncation can still happen. In that case the cursor is untrustworthy, so it is dropped and a new `truncated: true` field is set on `PageEnvelope` alongside `hasMore: true`, meaning "rows were dropped; resume position is unknown — restart or narrow the query."

## Related Issue

Closes #329

## Context

The SDK cannot fix this by computing a cursor. Cursor encoding is producer-owned and opaque — built-in domains use `base64url` of `{v,scope,index}`, third-party plugins may use anything — so there is no way to express "row 5 of this page". Hence the two-part approach: make truncation not happen in the common case, and signal it honestly when a producer forces it.

`hasMore` is deliberately `true` whenever `truncated` is set: we positively know unconsumed rows exist, since we just discarded them, even though the resume position is lost.

Reviewer notes:

- One existing test (`caps an oversized first page...`) encoded the bug as expected behavior, asserting that a limit-ignoring producer's stale cursor was propagated past a discarded row. Its expectations were updated. Worth a look, since the bug had a test blessing it.
- `merged.page.limit` still reports the producer's real page size, not the SDK's internally reduced request limit.
- `page.reset` handling is untouched and its existing tests pass unmodified.
- `validateAutoPagination` still requires `pagesLimit` alongside `maxItems`. That constraint is arguably now unnecessary, since `maxItems` alone is well-defined once requests are budgeted, but relaxing it is a public API change and is left out of a bugfix.
- Deliberately excludes CLI output changes. #322 renders `page.nextCursor` into a runnable `next` command, which is what makes this bug user-visible; consuming the new `truncated` signal there is a follow-up on that branch, since the code it needs does not exist on `main`.

## Testing

- Added regression tests to `packages/agent-sdk/src/__tests__/pagination.test.ts` for both loss paths, using a fake producer that mirrors the real cursor encoding in `local-domains.ts`. Each asserts on concrete row identities: the union of rows received plus rows reachable via the returned cursor. Both were confirmed failing before the fix (`expected [5,6,...,19] to deeply equal []` and `expected [7,8,9] to deeply equal []`).
- `npx turbo run typecheck lint test --force` across the whole repo: **99 successful, 99 total**.
- Version plan added: `.changeset/agent-sdk-max-items-row-loss.md` (`@rozenite/agent-sdk` patch, `@rozenite/agent-shared` minor for the new `PageEnvelope` field).

### DIFF
--- a/.changeset/agent-sdk-max-items-row-loss.md
+++ b/.changeset/agent-sdk-max-items-row-loss.md
@@ -1,0 +1,8 @@
+---
+"@rozenite/agent-sdk": patch
+"@rozenite/agent-shared": minor
+---
+
+Fix `autoPaginate.maxItems` silently dropping rows and reporting a `nextCursor` that skipped past them. The auto-pagination helper now requests exactly the rows still needed on every call (including the first), so a well-behaved tool's own cursor stays correct by construction instead of pointing past rows the caller never received.
+
+For tools that clamp or ignore the requested `limit` and still return more rows than asked, the extra rows are trimmed and the page's `nextCursor` is dropped rather than propagated (it can no longer be trusted to resume correctly). This is now signalled explicitly via a new `truncated: true` field on `PageEnvelope` (`@rozenite/agent-shared`), paired with `hasMore: true`, meaning "rows were dropped; resume position is unknown - restart pagination or narrow the query."
--- a/packages/agent-sdk/src/__tests__/pagination.test.ts
+++ b/packages/agent-sdk/src/__tests__/pagination.test.ts
@@ -1,6 +1,50 @@
 import { describe, expect, it, vi } from 'vitest';
 import { callToolWithOptionalPagination } from '../pagination.js';
 
+// Mirrors the real cursor encoding used by production paginated tools
+// (e.g. packages/middleware/src/agent/local-domains.ts `paginateRows`):
+// base64url JSON of { v: 1, scope, index }.
+const encodeCursor = (scope: string, index: number): string => {
+  return Buffer.from(
+    JSON.stringify({ v: 1, scope, index }),
+    'utf8',
+  ).toString('base64url');
+};
+
+const decodeCursorIndex = (cursor: string): number => {
+  const decoded = JSON.parse(
+    Buffer.from(cursor, 'base64url').toString('utf8'),
+  ) as { index: number };
+  return decoded.index;
+};
+
+// A well-behaved paginated tool: honors whatever `limit` it's asked for on
+// each call, and encodes `nextCursor` as the exact index to resume from.
+// Used to prove the row-loss is caused by the SDK's own request/merge
+// logic, not by a misbehaving producer.
+const createWellBehavedPaginatedTool = (
+  scope: string,
+  totalRows: number,
+) => {
+  const rows = Array.from({ length: totalRows }, (_, i) => ({ id: i }));
+  return vi.fn().mockImplementation((_name: string, args: unknown) => {
+    const record = (args ?? {}) as { limit?: number; cursor?: string };
+    const limit = record.limit ?? totalRows;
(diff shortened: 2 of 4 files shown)


V3RON:
Closing as won't do. We are not going to make the SDK a second pagination authority: SDK-side auto-pagination, maxItems, pagesLimit, response trimming, and the truncated signal will not be pursued. Pagination should be producer-owned, with a shared helper for plugin authors under #320; the SDK will pass a tool's page and continuation through unchanged.

PR #378, opened 2026-08-06
feat(ui): add List and NestedList components

V3RON:
## Description

Extracts the storage plugin's storages list into shared, reusable `@rozenite/ui` primitives:

- `List` — a flat, grouped, selectable list (`List.Group`, `List.Item`) with a new `adornment` slot for leading icons, alongside the existing `trailing` slot.
- `NestedList` — a collapsible tree (`NestedList.Item`) that nests to an arbitrary depth with automatic left indentation and expand/collapse (controlled or uncontrolled).
- `Sidebar` now composes `List` internally instead of duplicating item/group markup, so existing consumers (`storage-plugin`, `shell`, `controls-plugin`, `overlay-plugin`, `sqlite-plugin`) keep working unchanged while `Sidebar.Item` gains the `adornment` prop for free.
- `storage-plugin`'s storages list now passes a `Database` adornment icon per item (small usage update, no behavior change beyond the icon).
- Storybook stories added for `List` and `NestedList`, and the `Sidebar` story updated to demonstrate `adornment`.

Also includes one small, unrelated fix required to get this branch through the pre-push hook: an unused `vi` import in `packages/cli/src/__tests__/package-manager-detection.test.ts` (pre-existing on `main`, from a recent CLI commit, tripping the repo-wide `typecheck`).

## Related Issue

No related issue — this was scoped and requested directly in conversation, without an issue.

## Context

The storages list in `storage-plugin` was already using `@rozenite/ui`'s shared `Sidebar` component (confirmed via Storybook and source review) — it wasn't a duplicated/local component. What was missing was generality: `Sidebar` bakes in `<nav>` rail semantics and had no icon slot or nesting support, so a nested/tree use case (e.g. a future key browser) would have had to duplicate the item styling from scratch.

`List` and `NestedList` factor that styling into standalone, non-nav primitives that `Sidebar` now wraps. `Sidebar`'s public API is unchanged (same props, same rendered classes for the root), so no consumer migration is needed — `Sidebar.Group`/`Sidebar.Item` are now re-exports of `List.Group`/`List.Item`.

Confirmed the shell's plugin list already reuses `Sidebar.Item` at the same size/spacing as `storage-plugin`'s list, so there was no reason to introduce a second, larger variant for top-level navigation.

Per instructions, only `@rozenite/ui` and `@rozenite/storage-plugin` were touched — `sqlite-plugin` (and other `Sidebar` consumers) were left as-is since they need no changes to keep working.

## Testing

Automated:

- `pnpm turbo run lint typecheck test --filter=@rozenite/ui --filter=@rozenite/storage-plugin` — all pass (33 `@rozenite/ui` tests including new `NestedList` coverage for toggle/indentation/controlled-mode/leaf behavior; 193 `@rozenite/storage-plugin` tests)
- `pnpm release:plan` — changeset present and valid
- `pnpm format:all` — clean for changed files
- Full repo `pnpm lint:all` / `pnpm typecheck:all` pass (via the pre-push hook)

Manual:

- Ran Storybook (`pnpm start:ui-storybook`) and visually verified the new `List` and `NestedList` stories and the updated `Sidebar` story (adornment icons, badges, grouping, selection state).
- Verified `NestedList` collapse/expand behavior and per-depth indentation (8px/24px/40px) directly against the DOM in a live Storybook instance.


V3RON:
### Storybook screenshots

**List** — flat, grouped, selectable, with the new `adornment` icon slot and existing `trailing` badge slot:

List story

**NestedList** — collapsible, indefinitely nested tree with automatic per-depth left indentation, chevrons, and the same adornment/selection styling:

NestedList story

V3RON:
### Follow-up: horizontal scroll for deep nesting

Per discussion — `NestedList.Item` rows previously used `min-w-0`/`truncate`, which silently clipped long labels even when wrapped in `ScrollArea` (which already renders a horizontal scrollbar). Rows now use `min-w-full` instead, so they keep their natural width and genuinely overflow when content is wider than the viewport — `ScrollArea` then scrolls that overflow into view. Verified in Storybook: added a `DeepNestingWithHorizontalScroll` story and confirmed via the live DOM that a long, deeply-nested label becomes reachable by scrolling (`scrollWidth: 398` vs `clientWidth: 254`), with no regression to the normal case (full-row selection highlight still spans the row when content fits).

Also confirmed: `NestedList` intentionally has no `Group` (labeled section header) — only `List` supports that. Not adding it now since no current consumer needs labeled sections inside a tree.

PR #382, opened 2026-08-10
feat(cli): add rozenite skills command and flatten agent skill docs

V3RON:
## Description

Restructures Rozenite's agent skills so all real content is CLI-bundled and served by a new `rozenite skills` command, and the standalone installable skill becomes a thin router.

- Added `rozenite skills list` and `rozenite skills show <id>` to the CLI, backed by a small registry (`packages/cli/src/skills/registry.ts`) that reads bundled docs from `packages/cli/docs/` and exposes lookup by doc id and by domain token.
- Flattened all skill content into a namespace of short, self-contained docs bundled in `packages/cli/docs/`: `core` (shared ground truths), `cli` (shell-driven workflow), `sdk` (programmatic workflow), `sdk-patterns` (copy-pastable code patterns), and one doc per agent-enabled domain (`storage`, `mmkv`, `console`, `network`, `network-activity`, `react`, `react-navigation`, `redux-devtools`, `performance`, `memory`, `file-system`, `controls`, `tanstack-query`).
- Replaced the `rozenite-agent` and `rozenite-agent-sdk` installable skills with a single `rozenite` skill (`packages/cli/skills/rozenite/SKILL.md`) that only tells the agent to run `npx rozenite skills list`/`show <id>` and start with `core` — it never hardcodes the doc inventory, so it can't go stale.
- Updated `website/src/docs/agent/{skills,overview,sdk}.mdx` to document the new skill and command.
- Added a changeset for the `rozenite` package.

No MCP server or `rozenite mcp` command is included — that is a follow-up. No version-check behavior was touched.

## Related Issue

N/A — internal restructuring of existing agent-skill tooling.

## Context

The two previous skills (`rozenite-agent`, `rozenite-agent-sdk`) embedded all their content directly inside the installable skill package, including domain references and code-pattern references between files. That made the skill large, coupled its content to the `skills` CLI's installation format, and required cross-file references between skill documents. Moving the content into the CLI package and exposing it via `rozenite skills` means the docs can be read on demand (`npx rozenite skills show <id>`), each doc is self-contained (no cross-references), and the installed skill itself only needs to change if the discovery mechanism changes, not when doc content changes.

The registry resolves its docs directory the same way `TEMPLATE_DIR` resolves `template/` for `init` (relative to the built CLI entry point via `import.meta.url`), so it works from source and from the published package.

## Testing

Automated:

- `pnpm --filter rozenite run typecheck`
- `pnpm --filter rozenite run lint`
- `pnpm --filter rozenite run test` (94 tests passed, including new `skills-registry.test.ts` covering: all expected doc ids are listed, every domain doc has its `domain` field set (and non-domain docs don't), domain-token lookup works, unknown id/domain lookups return `undefined`, and no doc body contains a cross-reference to another doc's filename or a `references/` path)
- `pnpm lint:all`, `pnpm format:all`, `pnpm typecheck:all` (all packages pass)
- `pnpm release:plan` (confirms the changeset is picked up)

Manual:

- `node packages/cli/bin.js skills list` — prints all 17 docs with id/description (and `domain` where applicable)
- `node packages/cli/bin.js skills show core` — prints the real doc body
- `node packages/cli/bin.js skills show does-not-exist` — exits non-zero with a message listing valid ids


V3RON:
Addressed review feedback:

1. Removed the misleading "memoized per directory" claim from `loadAllDocs`'s doc comment in `packages/cli/src/skills/registry.ts` — it never memoized, and now the comment just says it reads from disk.
2. `rozenite skills list` now writes compact JSON by default with an opt-in `--pretty` flag, matching the `agent` command convention.
3. `core.md` now says `npx rozenite skills show cli` / `npx rozenite skills show sdk` instead of bare `cli`/`sdk` ids, so it's actionable without ambient knowledge of the fetch mechanism.
4. Extended the cross-reference test in `packages/cli/src/__tests__/skills-registry.test.ts` to reject doc ids adjacent to `docs/`, `domains/`, or `references/` markers (in addition to `<id>.md`). Also added a dedicated unit test for the matching helper itself, since a literal "any slash adjacent to a known id" rule produces false positives against `core`'s legitimate `evil/mmkv` / `avasapp/ably` domain-token-provenance examples and `sdk`'s `./sdk` npm subpath-export mentions — both are prose about naming conventions, not doc-location references.
5. `website/src/docs/agent/skills.mdx` now states plainly that the guidance lives in the `rozenite` CLI package and is versioned with it (so it can't drift from the installed Rozenite version), that the installed skill carries no workflow content of its own, and that `npx rozenite skills list`/`show <id>` work identically without installing the skill at all.

Re-ran `pnpm --filter rozenite run typecheck/lint/test` (95 tests pass), `pnpm lint:all`/`typecheck:all`, and the manual checks: `node packages/cli/bin.js skills list` (compact), `skills list --pretty`, `skills show core`, and `skills show does-not-exist` (non-zero exit, valid-id list).

V3RON:
## Smoke tested ✅

Ran the restructured skills + CLI end to end against a live app (iPhone 17 Pro simulator, playground app) via a Haiku sub-agent, using only the bundled docs (no prior Rozenite knowledge):

- **Skill discovery**: `npx rozenite skills list`, `skills show core`, `skills show cli` — all returned correct, readable content.
- **Live round trip**: agent read `skills show controls` to learn the domain's tool names, created a session (`agent session create`), toggled the `mockLatency` control in the simulator via `agent-device`, then called `agent controls call --tool get-item ...` and confirmed the CLI read back `value: true` — matching the on-device state.

Also verified separately: build + unit tests pass, `npm pack --dry-run` includes `docs/` and `skills/rozenite/SKILL.md`, and found & staged a fix for a dangling `.agents/skills/rozenite-agent` symlink left over from the old skill layout (staged locally, not yet pushed: points at `packages/cli/skills/rozenite`).

Confirms the relocated skill router + CLI-bundled docs are self-sufficient for a low-capability model to discover and drive correctly.

PR #386, opened 2026-08-12
feat: implement feature flags plugin

### LINKED ISSUE #383: Feature flags plugin
### Description

Most teams past a certain size end up building a homegrown debug screen for feature flags: seeing which flags are active on the current device and forcing a flag on/off to reproduce a bug or test a variant. There is no unified tooling for this in the React Native ecosystem, and it is not covered by React Native DevTools.

This is the same pain that the controls plugin solved for ad-hoc debug screens, applied to a narrower but very common domain.

### Suggested solution

A `@rozenite/feature-flags-plugin` following the adapter pattern established by the storage plugin:

- A generic adapter interface (list flags, get value, override, clear override) with first-party adapters for popular providers (e.g. LaunchDarkly, Statsig) plus a trivial adapter for local/custom flag implementations.
- Panel: list all flags with current values and sources (remote value vs. local override), search/filter, toggle/override, reset a single override or all of them.
- Overrides live on-device (in the adapter), so they survive DevTools disconnects but are dev-only; the hook is a no-op in production builds, like other plugins.
- Agent tools under the plugin domain: `list-flags`, `get-flag`, `override-flag`, `clear-overrides` — e.g. so an agent can reproduce a bug report that only occurs with a given flag enabled.

### Additional context

Proposed during a roadmap brainstorming session as one of the gaps that is achievable with the current, JS-only plugin architecture.

V3RON:
## Summary

Implements `@rozenite/feature-flags-plugin` following the adapter pattern established by the storage plugin. This addresses the need for unified debug tooling around feature flags in React Native.

## What's included

- **Adapter interface**: Generic pattern with first-party adapters for LaunchDarkly, Statsig, and a custom/local flags adapter
- **Feature flags panel**: UI to list all flags with current values and sources, search/filter, toggle overrides, and reset
- **Override store**: Device-side persistence that survives DevTools disconnects (dev-only)
- **Device hook**: Lightweight hook for accessing flags with RPC handlers
- **Agent tools**: `list-flags`, `get-flag`, `override-flag`, `clear-overrides` for agent-driven testing
- **Documentation**: Full docs with playground example and version plan

## Production readiness

- Hook is a no-op in production builds
- Hardened against malformed calls
- Proper lifecycle management to prevent RPC teardown on re-renders

Fixes #383

### DIFF
--- a/apps/playground/src/app/components/ui/KeyValueList.tsx
+++ b/apps/playground/src/app/components/ui/KeyValueList.tsx
@@ -1,0 +1,29 @@
+import { View } from 'react-native';
+import { List } from './List';
+import { KeyValueRow } from './KeyValueRow';
+import { useTheme } from '../../theme/useTheme';
+
+type KeyValueListItem = {
+  key?: string;
+  label: string;
+  value: string;
+};
+
+type KeyValueListProps = {
+  title?: string;
+  items: KeyValueListItem[];
+};
+
+export const KeyValueList = ({ title, items }: KeyValueListProps) => {
+  const { theme } = useTheme();
+
+  return (
+    <List title={title}>
+      {items.map((item) => (
+        <View key={item.key ?? item.label} style={{ paddingHorizontal: theme.spacing['2xl'] }}>
+          <KeyValueRow label={item.label} value={item.value} />
+        </View>
+      ))}
+    </List>
+  );
+};
--- a/apps/playground/src/app/components/ui/index.ts
+++ b/apps/playground/src/app/components/ui/index.ts
@@ -14,3 +14,4 @@ export { Badge } from './Badge';
 export type { BadgeVariant } from './Badge';
 export { EmptyState } from './EmptyState';
 export { KeyValueRow } from './KeyValueRow';
+export { KeyValueList } from './KeyValueList';
--- a/apps/playground/src/app/screens/ControlsPluginScreen.tsx
+++ b/apps/playground/src/app/screens/ControlsPluginScreen.tsx
@@ -5,7 +5,7 @@ import {
   Card,
   Field,
   Input,
-  KeyValueRow,
+  KeyValueList,
   PluginHeader,
   Row,
   Screen,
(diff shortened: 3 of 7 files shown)


V3RON:
## Agent tools QA (feature-flags-plugin)

Exercised all 4 agent tools (`list-flags`, `get-flag`, `override-flag`, `clear-overrides`) via the Rozenite CLI against the playground's custom adapter (`app` provider: boolean, string, number, json flags), both individually and in cooperation.

**Sequence run:**
1. `list-flags` (no args) — baseline, resolved sole provider, no overrides.
2. `get-flag` with explicit and omitted `providerId` — both resolved correctly.
3. `get-flag` on an unknown key — errored cleanly, no crash.
4. `override-flag` across all 4 types — each returned the updated flag with `overridden: true`.
5. `list-flags` — confirmed all 4 overrides visible together.
6. `override-flag` with a type-mismatched value (string for a number flag) — errored, no silent coercion.
7. `override-flag` on an unknown key — errored, nothing written.
8. `clear-overrides` with a single `key` — cleared just that flag, returned it reset to base value; other 3 stayed overridden.
9. `clear-overrides` with `providerId` only (no key) — cleared all remaining overrides.
10. `list-flags` — confirmed full reset to declared defaults.

No bugs found. Behavior matches the docs exactly: typo/unknown-key rejection, type validation, `providerId` auto-resolution to the sole provider, and per-key vs. whole-provider clearing all work as specified.

V3RON:
## Rebased due to duplicate merge from #387

#387 (pnpm v10→v11 migration) accidentally squash-merged an earlier snapshot of this branch's commits into `main` (everything through `09de6421`, "harden production entry stubs against malformed calls").

To make this branch mergeable again, I rebuilt it as a single commit on top of current `main`, containing only what wasn't already there — `667b85b7` ("polish DevTools panel UX"). Verified the resulting diff against `main` matches that commit exactly (no lost or duplicated changes), and confirmed lint, typecheck, tests, and build all pass.

No functional changes from what was already reviewed — this is purely a history fixup.

PR #403, opened 2026-08-13
feat(middleware): report changed prop/state/context keys in getRenderData

AndreiCalazans:
## Description

Enriches the built-in React profiler `getRenderData` agent tool so each rendered fiber reports **why** it re-rendered at key granularity, not just at category granularity. Each item now also carries:

- `displayName` — the resolved component name (previously only the numeric `fiberId` was returned, forcing a second round-trip to name a fiber), and
- `changedKeys` — the exact changed **prop / state / context key names**, plus `hooks` / `isFirstMount` flags, alongside the existing category-level `changeTypeHints`.

The middleware already parsed these names out of React DevTools change descriptions (`getChangeDescriptions` in `profiling-store.ts` keeps the `props`/`state`/`context` name arrays); they were simply collapsed to `changeTypeHints` and dropped before the response. This surfaces them.

### Example

Before, a slow commit could tell you a component rendered because of "props". Now it tells you which props:

```jsonc
{
  "fiberId": 34549,
  "displayName": "TokenRow",
  "actualDurationMs": 445.6,
  "selfDurationMs": 0.7,
  "isSlow": true,
  "changeTypeHints": ["props", "hooks"],
  "changedKeys": { "props": ["data", "isSelected"], "hooks": true }
}
```

### Changes

- `agent/runtime/react/types.ts`: add `ReactRenderDataChangedKeys`; add `displayName` and `changedKeys` to `ReactRenderDataItem`.
- `agent/runtime/react/store.ts`: extract `toChangeTypeHints` and new `toChangedKeys` to module scope (exported for testing) and emit `displayName` + `changedKeys` from `getRenderData`.
- `agent/local-domains.ts`: expose the new fields in the `getRenderData` pagination `fields`/`defaultFields` and expand the tool description.
- Add unit tests for `toChangeTypeHints` / `toChangedKeys`.
- Changeset: `@rozenite/middleware` minor.

The response shape is purely additive — existing consumers that read `fiberId`/`changeTypeHints` are unaffected. `changedKeys` is omitted when React recorded no change description for a fiber (e.g. the initial mount), matching how `changeTypeHints` behaves today.

## Related Issue

No existing issue. Happy to open one first if preferred per CONTRIBUTING — this is a small, additive enrichment of an existing agent tool and I wanted to show the concrete diff. This came out of driving `getRenderData` from a coding agent to answer "which components re-rendered and why" from the CLI.

## Context

This makes `getRenderData` self-sufficient for re-render analysis: an agent can attribute a slow commit to specific components and the specific props/state/context/hooks that invalidated them without a follow-up `getComponent`/`getNode` call per fiber.

## Testing

Automated (run in a partially-provisioned workspace; monorepo sibling packages `@rozenite/tools`/`@rozenite/agent-shared` are not built in my env, so the suites that import them fail to resolve on both this branch and `main` — unrelated to this change):

- `vitest --run src/__tests__/react-change-descriptions.test.ts` — 10 new tests pass.
- Full `@rozenite/middleware` suite: this branch = 40 passing / 5 files failing on the pre-existing sibling-package resolution error; `main` baseline = 30 passing / 6 files failing. This change only adds passing tests.
- Type-checking `store.ts` / `types.ts` / `local-domains.ts` surfaces no errors from this change (the only diagnostics are the same pre-existing sibling-module resolution errors present on `main`).
- `oxfmt --check` clean on all changed files.

Please run the full `pnpm --filter @rozenite/middleware typecheck build lint test` in CI / a complete workspace to confirm; I could not complete `pnpm install` against my environment's registry.

Manual: exercised the enriched tool end-to-end against a running React Native app (start profiling → navigate to force commits → read `getRenderData`) and confirmed `displayName` and `changedKeys` (including specific prop/state key names) appear for real updates and are absent for initial mounts.

### DIFF
--- a/.changeset/render-data-changed-keys.md
+++ b/.changeset/render-data-changed-keys.md
@@ -1,0 +1,5 @@
+---
+'@rozenite/middleware': minor
+---
+
+Enrich the `getRenderData` agent tool so each rendered fiber also reports its resolved `displayName` and a `changedKeys` object — the exact changed prop/state/context key names plus `hooks`/`isFirstMount` flags — alongside the existing category-level `changeTypeHints`. This lets a coding agent (or a human) see not just that a component re-rendered, but which specific props/state/context/hooks invalidated it, without a second round-trip to resolve fiber IDs to component names.
--- a/packages/cli/docs/react.md
+++ b/packages/cli/docs/react.md
@@ -29,6 +29,11 @@ Search and inspect:
 Profile:
 `startProfiling` -> reproduce interaction -> `stopProfiling` -> `getRenderData`.
 
+`getRenderData` rows carry `displayName` and timing by default. Add
+`--fields ...,changedKeys` for the exact changed prop, state, and context key
+names behind `changeTypeHints`, which attributes a re-render without a
+follow-up `getComponent` call per fiber.
+
 CLI discovery listings and paginated React calls (`getTree`, `searchNodes`,
 `getChildren`, `getProps`, `getState`, `getHooks`, and `getRenderData`) use
 columnar `cols` and `rows` output when two or more rows are returned. Their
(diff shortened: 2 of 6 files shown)


V3RON (approved):
Looks good! Thanks!

PR #408, opened 2026-08-13
fix(runtime,middleware): stop dropping host-to-device messages with astral-plane characters

### LINKED ISSUE #407: bug: host-to-device messages containing emoji/astral characters are silently dropped
## Description

Rozenite sends host-to-device messages (`packages/runtime/src/rn-devtools/bindings-model.ts`'s `sendMessage`, and `packages/middleware/src/agent/session.ts`'s `sendDomainMessage`) by JSON-stringifying the payload and interpolating the escaped string into a `Runtime.evaluate` JS source-string expression, which the device's JS engine (Hermes) then has to recompile.

Any message whose payload contains a character outside the Basic Multilingual Plane (for example an emoji, a UTF-16 surrogate pair) fails that recompilation — Hermes reports `Invalid UTF-8 code point` — and the message is silently lost. Neither call site inspects the `Runtime.evaluate` response for `exceptionDetails`, so the failure is invisible; the caller believes the send succeeded.

## Reproduction

1. Have a plugin send a host→device message whose payload contains an emoji (e.g. `storage-plugin`'s `set-entry` with a value containing 🎉).
2. Observe that the message never reaches the device — no error is surfaced anywhere.
3. Confirmed via a raw CDP script: `Runtime.evaluate` with a double-JSON-stringified emoji-bearing expression fails with `Compiling JS failed: ... Invalid UTF-8 code point 0xd83c`; the identical payload via `Runtime.callFunctionOn` with a by-value argument round-trips byte-identical.

## Impact

Any host→device message carrying user data can be affected — `storage-plugin`/`mmkv-plugin` `set-entry`, `sqlite-plugin` query values, `network-activity-plugin` overrides, and the agent bridge's own `agent-session-ready` handshake message.

## Suggested fix

Switch the send path to `Runtime.callFunctionOn` with the payload passed as a single-stringified by-value argument (rather than double-stringified into JS source text), and check the response for failure instead of ignoring it.

V3RON:
## Description

Host-to-device messages (plugin domain messages, the agent-session-ready handshake, `set-entry` writes for storage/mmkv/sqlite plugins, etc.) were sent by double-JSON-stringifying the payload and interpolating the result into a `Runtime.evaluate` JS source-string expression, which the device's JS engine (Hermes) then recompiles.

A payload containing a character outside the Basic Multilingual Plane (for example an emoji, a UTF-16 surrogate pair) fails that recompilation with `Invalid UTF-8 code point`, and the message is silently lost — neither call site checked the response for `exceptionDetails`, so the failure was invisible.

This switches the send path to `Runtime.callFunctionOn` with the payload passed as a single-stringified by-value argument (in both `packages/runtime/src/rn-devtools/bindings-model.ts`, the DevTools-frontend model, and `packages/middleware/src/agent/session.ts`, the middleware's raw-CDP agent session), and stops silently swallowing failures.

## Related Issue

Closes #407

## Context

**Why `Runtime.callFunctionOn`, and did the frontend wrapper actually expose it?** The CDP protocol supports `callFunctionOn`, but the DevTools-frontend's hand-written `RuntimeAgent` ambient declaration (`rn-devtools-frontend-api.d.ts`) only declared `invoke_evaluate`/`invoke_addBinding`. Verified directly against the installed `@react-native/debugger-frontend@0.86.2` bundle: `core/sdk/sdk.js` both registers `Runtime.callFunctionOn` (via `protocol_client.js`'s command registration) and already calls `agent.invoke_callFunctionOn(...)` internally (in `RemoteObject`'s property-mutation helpers) — so it's exposed, and both `bindings-model.ts` and `session.ts` implement the switch.

**Correctness findings surfaced during an adversarial review (opus) of the first pass, all fixed:**
- The real `RuntimeAgent`'s generated `invoke_*` methods never reject — a protocol-level failure (e.g. a stale execution context id after a reload) resolves with `getError()` set rather than throwing. The first pass only checked `exceptionDetails` (a *different* failure mode: a JS-level exception during the evaluated function) and had a dead `.catch()`. Both are now checked.
- The pending-command timeout (added because a malformed frame can come back as an uncorrelatable `"id": null"` response) was initially applied to every CDP command, which would have cut off legitimately long-running commands (`HeapProfiler.takeHeapSnapshot`, large `Network.getResponseBody`/`getRequestPostData` calls). It's now opt-in per command and only applied to the send path.
- `sendDomainMessage` initially always returned an already-resolved promise once the round trip was no longer awaited, which silently broke `bootstrap()`'s existing retry-on-failure logic for the `agent-session-ready` handshake. It now returns the real (still-rejectable) promise chain, so `void`-calling sites stay non-blocking while the one `await`-ing call site keeps its retry behavior.
- The unpaired-surrogate sanitizer initially scanned for raw UTF-16 surrogate code units, but by the time it ran (on `JSON.stringify(message)`'s output) a lone surrogate is already represented as a literal `\uXXXX` *escape sequence* (`JSON.stringify` is well-formed since ES2019 and never emits a raw lone surrogate) — so the original sanitizer never matched anything. It now matches the escape-sequence text instead, verified against a real device: the device's own JSON parser rejects an unpaired `\uXXXX` escape sequence with `"expected another unicode escape for second half of surrogate pair"`, even though it's syntactically valid JSON.

**Honest framing:** this is not a meaningful throughput/perf win. Median added latency from the extra `Runtime.callFunctionOn` framing vs `Runtime.evaluate` is 0–1ms for typical (sub-8KB) structured payloads on a localhost simulator connection; with unstructured (quote-free) payloads the new framing is marginally *larger* on the wire. The actual win is removing the round-trip await from the hot send path (CDP requests are ordered and execute in order regardless of whether the caller awaits the response) and, primarily, fixing the data loss.

**⚠️ Known overlap with #405 (`perf/skip-json-parse-non-rozenite-bindings`), both open off `main`:** both PRs independently add the same line to `packages/runtime/package.json`'s `scripts`:
```
"test": "vitest --run --passWithNoTests"
```
Git will conflict on that hunk for whichever merges second. The two PRs also disagree on convention: #405 adds the script *without* declaring `vitest` as a devDependency, on the assumption that packages here don't declare it directly. That assumption doesn't hold — `packages/middleware`, `shell`, `metro`, `tools`, `expo-atlas-plugin`, `file-system-plugin`, and `sqlite-plugin` all declare `vitest` in `devDependencies`. This PR follows that majority convention (adds `"vitest": "^4.0.18"` to `packages/runtime`'s `devDependencies`, with the corresponding minimal `pnpm-lock.yaml` update). Whoever merges second should rebase onto the other and drop the duplicate script line; if #405 lands first, this PR should additionally pick up its devDependency declaration rather than the reverse.

## Testing

Automated:
- `pnpm --filter @rozenite/runtime typecheck / lint / test` — 15 tests passed
- `pnpm --filter @rozenite/middleware typecheck / lint / test` — 107 tests passed
- `pnpm --filter @rozenite/metro test` (CJS-load smoke test over the built `@rozenite/runtime`/`@rozenite/middleware` bundles) — 2 tests passed
- `pnpm release:plan` — changeset present and valid
- `pnpm build:all`, root `lint`/`typecheck` (also run automatically by the repo's pre-push hook) — all green

Manual, against a real device (RN 0.86 / Hermes, iPhone 17 Pro simulator, connected over the raw CDP `Runtime` domain with `Origin: http://127.0.0.1:8081`):
- Emoji-bearing payload via the new `Runtime.callFunctionOn` shape: delivered byte-for-byte (round-trip identical). The same payload via the old `Runtime.evaluate` double-stringify path: `Compiling JS failed: ... Invalid UTF-8 code point 0xd83c` (reproduces the bug for comparison in the same session).
- A message containing an unpaired UTF-16 surrogate, sanitized: delivered successfully with no `exceptionDetails`/error and a byte-identical round trip.

### DIFF
--- a/.changeset/fix-host-to-device-message-encoding.md
+++ b/.changeset/fix-host-to-device-message-encoding.md
@@ -1,0 +1,15 @@
+---
+'@rozenite/runtime': patch
+'@rozenite/middleware': patch
+---
+
+Fix host-to-device messages being silently dropped when their payload
+contained a character outside the Basic Multilingual Plane (for example an
+emoji). Messages were sent by interpolating a JSON string into a JS source
+string that the device's JS engine had to recompile, which cannot represent
+such characters and failed the recompilation without surfacing an error —
+losing the message. Sends now go through `Runtime.callFunctionOn` with the
+payload passed as a value argument instead, and a failed send is now reported
+instead of being silently swallowed. Sends also no longer wait for the device
+to acknowledge the round trip, since delivery order is already guaranteed by
+the underlying protocol.
--- a/packages/middleware/src/__tests__/agent-session.test.ts
+++ b/packages/middleware/src/__tests__/agent-session.test.ts
@@ -19,6 +19,7 @@ const mocks = vi.hoisted(() => {
   const loggerInfo = vi.fn();
   const loggerWarn = vi.fn();
   const loggerDebug = vi.fn();
+  const loggerError = vi.fn();
   const handler = {
     connectDevice: vi.fn(),
     disconnectDevice: vi.fn(),
@@ -38,6 +39,10 @@ const mocks = vi.hoisted(() => {
   const wsInstances: MockWebSocket[] = [];
   let bindingName = 'rozenite-binding';
   let stalledRuntimeExpression: string | null = null;
+  let stalledCallFunctionOn = false;
+  let forcedCallFunctionOnExceptionText: string | null = null;
+  let forcedCallFunctionOnProtocolError: string | null = null;
+  let forcedCallFunctionOnMessageError: string | null = null;
 
   class MockWebSocket {
     static readonly CONNECTING = 0;
@@ -95,6 +100,12 @@ const mocks = vi.hoisted(() => {
         method: payload.method,
         params: payload.params,
       });
+
+      if (payload.method === 'Runtime.callFunctionOn' && forcedCallFunctionOnProtocolError) {
+        callback?.(new Error(forcedCallFunctionOnProtocolError));
+        return;
+      }
+
       callback?.();
 
       if (
@@ -105,7 +116,25 @@ const mocks = vi.hoisted(() => {
         return;
       }
 
+      if (payload.method === 'Runtime.callFunctionOn' && stalledCallFunctionOn) {
+        return;
+      }
+
       queueMicrotask(() => {
+        if (payload.method === 'Runtime.callFunctionOn' && forcedCallFunctionOnMessageError) {
+          // A genuine CDP protocol-level error response
+          // (`{"id": ..., "error": {...}}`), as opposed to a socket-write failure
+          // or a successful response carrying `exceptionDetails`.
+          this.emit(
+            'message',
+            JSON.stringify({
+              id: payload.id,
+              error: { message: forcedCallFunctionOnMessageError, code: -32000 },
+            }),
+          );
+          return;
+        }
+
         this.emit(
           'message',
           JSON.stringify({
@@ -128,6 +157,13 @@ const mocks = vi.hoisted(() => {
   }
 
   const getCommandResult = (method: string, params?: Record<string, unknown>) => {
+    if (method === 'Runtime.callFunctionOn' && forcedCallFunctionOnExceptionText) {
+      return {
+        result: {},
+        exceptionDetails: { text: forcedCallFunctionOnExceptionText },
+      };
+    }
+
     if (method !== 'Runtime.evaluate') {
       return {};
     }
@@ -158,6 +194,7 @@ const mocks = vi.hoisted(() => {
     loggerInfo,
     loggerWarn,
     loggerDebug,
+    loggerError,
     MockWebSocket,
     handler,
     createAgentMessageHandler: vi.fn(() => handler),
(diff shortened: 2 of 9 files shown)


V3RON:
Closing: the change doesn't justify its scope.

This PR was motivated by removing the double-`JSON.stringify` on the host→device send path. Measurement doesn't support that as a win:

- End-to-end (median of 40 round trips, simulator over localhost): 0–1ms difference for payloads under 8KB, ~6ms at 52KB. With unstructured/quote-free payloads the new framing is marginally *larger* on the wire.
- The second `JSON.stringify` in isolation costs ~70µs on a 52KB payload and ~1.5µs at 1KB — and it runs on the host (DevTools page / Node), never on the device's JS thread.
- Host→device traffic is sparse and user-initiated (clicks, mount-time fetches, single edits), so none of it compounds.

That's a lot of architectural change — `Runtime.callFunctionOn`, an `executionContextId` lifecycle, and an extension to the hand-written frontend `.d.ts` — for no measurable gain.

The underlying bug (#407) is real and stays open. It can be fixed far more cheaply by ASCII-escaping non-ASCII code units before the JSON is embedded as JS source, which was verified against a device to fix it completely with no `callFunctionOn` and no perf claim attached.

Two findings from the review here are worth preserving regardless:
1. The DevTools frontend's generated `invoke_*` methods never reject — protocol errors surface only via `getError()`, so error handling written against `.catch()`/`exceptionDetails` alone is dead code.
2. By the time a sanitizer sees `JSON.stringify` output, a lone surrogate is already `\uXXXX` escape *text*, not a raw code unit — scanning for raw surrogate code units matches nothing.

PR #419, opened 2026-08-18
feat(ui): add tone tokens for the design system (Phase 0)

V3RON:
## Description

Adds the token layer for the `@rozenite/ui` design system roadmap (Phase 0 of `docs/agents/ui-design-system-plan.md`):

- New `success`, `warning`, and `info` tone tokens (`--<tone>`, `--<tone>-foreground`, `--<tone>-soft`) in `styles.css`, defined for both light and dark.
- A `--danger`/`--danger-foreground` alias for the existing `--destructive` pair (the eventual rename target for Phase 3).
- `src/tokens/tone.ts` exporting the `Tone` union.
- `src/tokens/size.ts` exporting the `Size` union and the height/padding/icon-size maps.
- `src/tokens/tone-variants.ts` exporting the `surfaceTone` and `textTone` cva recipes.
- A new Storybook "Foundations/Tokens" page rendering the full tone × variant matrix in both themes and documenting the square-by-design `--radius: 0` decision.

Purely additive — no existing component's className output or behavior changes.

## Related Issue

## Context

This is the first of several phase-scoped PRs implementing the plan. Phase 0 exists because nothing later in the plan (new primitives, tone/emphasis split on Button/Badge/IndicatorDot) can be built cleanly until tone and size are real, exported tokens instead of per-component inventions — today only `destructive` exists as a tone, which is why several plugin helpers hand-roll `text-green-400` etc.

While building the tokens Storybook page, the accessibility panel initially flagged 20 color-contrast violations. Two real issues surfaced and were fixed: the `soft` variant was pairing white foreground text on a pale tint (nearly invisible), and the initial 400-level tone colors were too light for legible text/border use in light mode. Colors were adjusted so light mode uses darker, more saturated tones (matching the existing pattern of darker-for-light / lighter-for-dark), bringing violations down to a residual few borderline/pre-existing ones on this documentation-only page (the a11y check here is informational, `test: 'todo'`, non-blocking).

## Testing

- Manual: ran `pnpm start:ui-storybook`, opened the new Foundations/Tokens story with `agent-browser`, and visually verified the tone × variant matrix in both light and dark panels, iterating on colors to reduce axe color-contrast violations.

### DIFF
--- a/.changeset/ui-design-system-phase-0-tokens.md
+++ b/.changeset/ui-design-system-phase-0-tokens.md
@@ -1,0 +1,5 @@
+---
+'@rozenite/ui': minor
+---
+
+Add tone tokens for `success`, `warning`, and `info`, plus a `--danger` alias for the existing `--destructive` pair. Export the `Tone` and `Size` unions and the `surfaceTone`/`textTone` cva recipes so components can compose consistent tone-aware styling. No existing component's behavior changes.
--- a/apps/ui-storybook/src/stories/Tokens.stories.tsx
+++ b/apps/ui-storybook/src/stories/Tokens.stories.tsx
@@ -1,5 +1,5 @@
 import type { Meta, StoryObj } from '@storybook/react-vite';
-import { surfaceTone, textTone, type Tone } from '@rozenite/ui';
+import { surfaceTone, textTone, sizeHeight, sizeIcon, type Tone, type Size } from '@rozenite/ui';
 
 const TONES: Tone[] = ['neutral', 'primary', 'success', 'warning', 'danger', 'info'];
 const VARIANTS = ['solid', 'soft', 'outline', 'ghost'] as const;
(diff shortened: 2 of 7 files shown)


V3RON:
Ran an independent Opus review against this diff. Findings and fixes, in 6a6ae795:

- Added a `--danger-soft` token so `danger` has the same soft-surface support as success/warning/info, instead of a hardcoded `bg-danger/10` (the token layer was asymmetric — `Alert`/`Badge` in later phases couldn't treat every tone uniformly).
- Bound `surfaceTone`/`textTone`'s tone keys to the `Tone` union via `satisfies Record<Tone, string>`, so the recipes can't silently drift from the locked union.
- Added a `SizeScale` story — the tokens page's own description claimed a "tone × variant × size matrix" but only rendered tone × variant.
- Dropped a redundant `withPluginShell: false` already set at the story's meta level.

Left `--danger` itself equal to `--destructive` (2.74:1 contrast in light mode) rather than diverging it, per the plan's explicit "Add `--danger` ... as aliases" instruction — Phase 3 is where the actual color/rename work happens; fixing it here would mean another jump later.

PR #420, opened 2026-08-18
feat(ui): add seven design system primitives (Phase 1)

V3RON:
## Description

Stacked on #419 (Phase 0 tokens). Adds Phase 1 of `docs/agents/ui-design-system-plan.md`: seven new `@rozenite/ui` primitives, each in its own name-matched folder with a Storybook story.

- `Text` + `Heading` — collapses ad-hoc typography strings into `title` / `heading` / `body` / `caption` / `code` / `numeric` variants; `Heading` renders semantic `h1`-`h6`.
- `Row` / `Column` — flex layout primitives with `gap`, `align`, `justify`, `wrap`, and the `fill`+`scroll` idiom (`min-{w,h}-0 flex-1 overflow-auto`) for scrollable panes, currently re-derived at 14 sites across 9 packages.
- `Icon` — renders any icon component at a size from the shared `Size` scale, plus a curated `lucide-react` re-export.
- `IconButton` — a square, icon-only button with a **required** `label` prop (an unlabelled icon button is a type error) that doubles as the accessible name and its own tooltip.
- `Menu` — built on Base UI's `Menu` primitive; a genuine gap today (`network-activity` reaches for raw Radix + `@floating-ui/react` because nothing exists).
- `Alert` — a tone-aware strip built on `surfaceTone` from Phase 0, replacing ad-hoc rose-toned strips.
- `Kbd` — a styled keyboard key/shortcut hint.

All additive — no existing component's behavior changes.

## Related Issue

## Context

Building each primitive in Storybook and testing interactively with `agent-browser` surfaced two real bugs, both fixed here:

- The `Menu` story initially placed `Menu.GroupLabel` as a sibling of `Menu.Group` instead of nested inside it, which Base UI rejects at runtime (`MenuGroupContext is missing`) — fixed by nesting it correctly, which is also the only valid usage, so this only affected the story, not the component.
- `Row`/`Column` with `scroll` rendered a `div` with `overflow-auto` but no way to reach it by keyboard, tripping axe's `scrollable-region-focusable` rule. Fixed by giving the scrollable variant `tabIndex={0}` by default (still overridable via an explicit `tabIndex` prop).

## Testing

- Manual: ran `pnpm start:ui-storybook` and drove each new story with `agent-browser` — opened the `Menu` dropdown, hovered an `IconButton` to confirm its tooltip and accessible name, checked `Alert` tones render with sufficient contrast in the ambient plugin theme, and confirmed the `Column`/`Row` `FillAndScroll` stories go from 1 accessibility violation to 0 after the `tabIndex` fix.

### DIFF
--- a/.changeset/ui-design-system-phase-1-primitives.md
+++ b/.changeset/ui-design-system-phase-1-primitives.md
@@ -1,0 +1,5 @@
+---
+'@rozenite/ui': minor
+---
+
+Add seven new `@rozenite/ui` primitives: `Text`/`Heading` for typography, `Row`/`Column` for flex layout (including the `fill`+`scroll` idiom for scrollable panes), `Icon` for rendering any icon component at a shared size scale plus a curated lucide re-export, `IconButton` for a labelled, tooltip-carrying icon-only button, `Menu` for dropdown actions, `Alert` for tone-aware inline status strips, and `Kbd` for keyboard shortcut hints. All additive — no existing component's behavior changes.
--- a/apps/ui-storybook/src/stories/Alert.stories.tsx
+++ b/apps/ui-storybook/src/stories/Alert.stories.tsx
@@ -1,0 +1,46 @@
+import type { Meta, StoryObj } from '@storybook/react-vite';
+import { Alert } from '@rozenite/ui';
+
+const meta = {
+  component: Alert,
+  title: 'Components/Alert',
+  parameters: {
+    docs: {
+      description: {
+        component: 'A tone-aware strip for surfacing status, warnings, or errors inline.',
+      },
+    },
+  },
+} satisfies Meta<typeof Alert>;
+export default meta;
+type Story = StoryObj<typeof meta>;
+
+/**
+ * Compare every tone.
+ * @summary Compare every alert tone.
+ */
+export const AllTones: Story = {
+  render: () => (
+    <div className="flex flex-col gap-2">
+      <Alert tone="neutral">
+        <Alert.Description>A neutral, informational note.</Alert.Description>
+      </Alert>
+      <Alert tone="info">
+        <Alert.Title>Heads up</Alert.Title>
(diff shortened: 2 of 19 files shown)


V3RON:
Ran an independent Opus review against this diff. Findings and fixes, in dd80fa3d:

- `Alert`: `Title`/`Description` were rendering side-by-side instead of stacked (the root was `flex items-start`, not `flex-col`); also switched to `role="status"` for non-danger tones instead of always `role="alert"` — an assertive live region for a static informational strip is announcement noise.
- `Menu`: the popup had no max-height, so a long menu would grow past the viewport instead of scrolling, unlike `Select`/`Combobox`; capped it to `--available-height` and added `alignOffset` for parity with `Tooltip`.
- `IconButton`: `size="sm"`/`"lg"` changed the button's box but not the glyph inside it (`buttonVariants({size:'icon'})` bakes in a fixed `[&_svg]:size-4`), so the icon looked identical at every size — fixed by pairing each size step with its own icon-size class; also switched the local size union to the shared `Size` type.
- `Text`/`Heading`/`Icon`: added a `tone` prop backed by `textTone` — Phase 0 built `textTone` explicitly for these three components and nothing consumed it yet. `Heading` levels now carry a real type-size scale (`h1` vs `h6` previously looked identical).
- `Icon`: defaults to `aria-hidden` unless given an `aria-label`, since the common case (e.g. inside `IconButton`, which already labels itself) is decorative.
- `Row`/`Column`: the `scroll` variant is now a tab stop (per the Phase 0 PR's fix) but had no visible focus ring — added one.
- `Kbd`: replaced an arbitrary `text-[10px]` bracket value with `text-xs`, since Phase 5 plans to ban arbitrary values.

Verified interactively with `agent-browser`: Alert now stacks correctly, IconButton glyphs visibly scale across sm/md/lg, and Menu still opens/closes correctly after the max-height change.

PR #421, opened 2026-08-18
feat(ui): add polymorphism and compound fixes to the design system (Phase 2)

V3RON:
## Description

Stacked on #420 (Phase 1 primitives). Adds Phase 2 of `docs/agents/ui-design-system-plan.md`: polymorphism and internal compound-component repairs, all additive/non-breaking (existing stories render unchanged).

1. **`render` polymorphism** — via Base UI's `useRender`, added to `Button`, `Badge`, `Link`, `List.Item`, `Card`, `EmptyState`, `PluginHeader.*`, and the Phase 1 primitives (`Text`, `Heading`, `Row`, `Column`, `IconButton`, `Alert`, `Kbd`). `Button` had no ref forwarding before this, so `Dialog.Close render={<Button/>}` worked only by accident of prop spreading — it now composes correctly.
2. **`Toast` is callable** — it was a plain object literal (`{ Provider: ... }`), unlike every other `Object.assign` namespace in the package. `Toast.Provider` remains as a deprecated alias.
3. **`Sidebar` gets its own slots** — `Sidebar.Group`/`Sidebar.Item` now emit `data-slot="sidebar-group"`/`"sidebar-item"` instead of aliasing `List.Group`/`List.Item` and emitting `data-slot="list-item"`.
4. **Label override props** for localization: `Dialog.Content`'s `closeLabel`, `SearchField`'s `clearLabel`, `Toast`'s `dismissLabel`, `Card`'s `collapseLabel`/`expandLabel`, `PluginHeader.ThemeSwitcher`'s `lightLabel`/`darkLabel`.
5. **Extracted duplicated recipes** — `fieldSurface` (6 sites: `Input`, `Textarea`, `SearchField`, `Field.Control`, `Select.Trigger`, `Combobox.Input`), `popoverSurface` and `optionRow` (2 sites each: `Select`, `Combobox`), and `SortIcon` (`DataTable`, `VirtualizedDataTable`) into `src/utils/`.
6. **Repaired `VirtualizedDataTable`** — fixed a real bug where its `className`/`scrollClassName` props replaced the base styling instead of extending it (they were plain default-string props, never merged through `cn`). Moved it into a name-matched folder. Aligned its empty-state copy and accessibility attributes with `DataTable`.

## Related Issue

## Context

Item 3 (Sidebar's own slots) duplicates `List.Item`'s implementation into `Sidebar.Item` rather than sharing it, since `List.Item` is itself styled with `--sidebar-*` tokens (it was apparently designed for sidebar use originally) — deduplicating that further wasn't in scope for this phase and isn't called out in the plan.

## Testing

- Manual: ran `pnpm start:ui-storybook` and drove the affected stories with `agent-browser` after each task — confirmed `Button`'s new `render` prop produces a real `<a href>` styled as a button, `Dialog.Close render={<Button/>}` (the existing `ConfirmDialog` story) still opens/focuses/closes correctly, `Toast`'s new callable form still shows and dismisses toasts, `Sidebar` renders unchanged, and `Input`/`Textarea`/`SearchField`/`Select`/`Combobox`/`DataTable`/`VirtualizedDataTable` all render unchanged after the recipe extraction and repair.


V3RON:
Ran an independent Opus review against this diff (in an isolated worktree, so it didn't disturb the branch). Two real bugs and a few consistency issues surfaced, fixed in 6f95a14c:

- **Real bug**: `EmptyState` and `PluginHeader.ThemeSwitcher` compose their own `children` internally (title/description/action, and the Sun/Moon icon) but didn't destructure `children` out of their props before spreading `...props` — since neither prop type excludes `children`, a caller passing `children` silently clobbered the composed content. Every sibling that does this (`Button`, `Link`, `List.Item`, `Sidebar.Item`) already destructures it correctly; these two didn't. Fixed by discarding `children` explicitly.
- **Real bug**: `DataTable` and `VirtualizedDataTable` were emitting `aria-sort="none"` on every header, including non-sortable ones — ARIA requires the attribute be *omitted* on headers that can't be sorted, not set to `"none"`. Gated both on `canSort`.
- **Real bug**: `VirtualizedDataTable`'s `style` prop had the exact same default-prop bug task 6 was meant to fix for `className`/`scrollClassName` — `style = { height: 400 }` as a default meant a caller passing `style={{ width: '100%' }}` lost the height entirely. Fixed with the same merge approach, plus a stable default object reference.
- Hoisted `Card`'s `useRender()` call out of a JSX child expression into a statement, matching every sibling.
- Added a `ToastProps` alias, since `ToastProviderProps` reads oddly now that the root itself is callable.
- Added story coverage for two things this PR's own "story per behavior change" rule was missing: `List.Item`'s new `render` prop (as an anchor) and a localized `Card` collapse toggle.

The review also flagged that `Sidebar.Group`/`Sidebar.Item` fully duplicate `List.Group`/`List.Item`'s implementation rather than sharing a recipe, which is a fair callout but was a deliberate choice within this phase's scope (`List.Item` is itself styled with `--sidebar-*` tokens, and the plan's task 3 only asked for correct `data-slot`s, not a new shared recipe) — left as-is to avoid scope creep, noted for anyone picking up further consolidation later.

PR #423, opened 2026-08-18
feat(ui)!: the breaking batch — size scale, tone/variant split, danger rename (Phase 3)

V3RON:
## Summary

Phase 3 of the `@rozenite/ui` design system plan (see `docs/agents/ui-design-system-plan.md`, not committed) — "the breaking batch." Stacked on #421 (Phase 2).

- `adornment` → `leading`/`trailing` on Button, List.Item, NestedList.Item, Sidebar.Item
- One `sm | md | lg` size scale across Button, Input, Textarea, Select.Trigger, Combobox.Input, SearchField, Field.Control, Toolbar.Button, Tabs, List.Item, NestedList.Item, Sidebar.Item, Badge — replacing inconsistent hardcoded heights (h-8/h-7/h-6 mismatches). `size="icon"` removed from Button in favor of `IconButton`.
- `--destructive` → `--danger` (canonical, no longer an alias); every `text-destructive`/`bg-destructive`/`border-destructive` utility renamed across the 8 consumers.
- Tone split from emphasis on Button, Badge, IconButton (`tone` + `variant: 'solid'|'soft'|'outline'|'ghost'`, reusing the `surfaceTone` recipe from Phase 0). IndicatorDot gains the full tone scale + a `size` prop.
- ConfirmDialog's `destructive: boolean` → `tone`; now accepts `className`.
- Removed layout-leakage margins: Toolbar.Separator's `mx-1` (spacing now comes from Toolbar's own `gap-1`), PluginHeader.Actions' `ml-auto` (PluginHeader root now uses `justify-between` — group leading content in one wrapper), Sidebar.Footer's `mt-auto` (redundant once the scrollable middle section uses `flex-1`).
- Loosened self-sizing: Sidebar's `w-56` is no longer a default — callers size it explicitly via `className`. **PluginShell's `h-screen` and EmptyState's `flex-1` were evaluated for the same treatment but kept as defaults after real consumer evidence showed dropping them breaks panel layout** (some consumer packages don't reset `html`/`body`/`#root` height, so a relative `h-full` collapses the panel to zero height; `EmptyState`'s `flex-1` is inert in non-flex parents so there's no downside to keeping it — see the correction commit for the full rationale). Toast gains a `viewportClassName` escape hatch to reposition its viewport.
- Migrated all 8 adopting consumer packages (storage, shell, react-navigation, vite-plugin, feature-flags, rhf, overlay, controls) to the new API, including moving `Button size="icon"` call sites to `IconButton` and passing explicit width to `Sidebar` where it relied on the removed default.

Full before/after migration table is in the changeset.

## Test plan

- [x] `pnpm --filter @rozenite/ui typecheck/lint/build/test` (33 tests)
- [x] `cd apps/ui-storybook && npx tsc -b --force` / `pnpm --filter ui-storybook lint`
- [x] Visual check of Button/Badge/IconButton/ConfirmDialog/Sidebar/PluginHeader/EmptyState stories via Storybook (tone colors, sizing, layout all match prior defaults)
- [x] `pnpm --filter <pkg> typecheck/lint` for all 8 adopting consumer packages
- [x] Full monorepo `pnpm lint:all` (62/62), `pnpm typecheck:all` (63/63), `pnpm turbo run test` (54/54) — all green


V3RON:
## Opus review — findings and fixes

Spawned an Opus review of the full Phase 3 diff, including the 8-consumer-package codemod and the PluginShell/EmptyState default-sizing correction. It found 8 issues; here's what I fixed and why, and what I declined.

**Fixed (real bugs):**
- **Shell's collapsed sidebar footer floated to the top.** `Sidebar.Footer` lost `mt-auto` in this PR, and Shell's collapsed rail has no `flex-1` scroll section to compensate (only the expanded state does) — so the two icon buttons ended up right under the logo instead of pinned to the bottom. Restored `mt-auto` on that call site.
- **feature-flags-plugin's reset button rendered purple instead of neutral.** Its `IconButton` was missing `tone="neutral"` — the only ghost/outline call site across all 8 packages that missed it (verified against 20+ others). Also dropped a redundant `title` now that `IconButton`'s `label` already provides a tooltip.
- **overlay-plugin's `styles.css` referenced the removed `--destructive` var**, leaving `--color-danger` unresolvable there — the vite-plugin equivalent was migrated but this one was missed. Pointed it at `--danger`.
- **`surfaceTone`'s hover states leaked into `Badge`.** I'd added hover states to `solid`/`outline` in the review-fix-before-this-one, correctly excluding `soft` because `Alert` (non-interactive) consumes it — but missed that `Badge` also consumes `solid`/`outline` and is just as non-interactive. Every badge was picking up a hover background. Fixed by making hover opt-in via a new `interactive: true` variant on `surfaceTone`, used by `Button`/`IconButton` only.
- **`Toolbar.Button` icon-only buttons stopped being square.** Several consumers sized them `w-7` to match the old `h-7` toolbar-button default; now that the default is `size="sm"` (`h-6`), those needed `w-6`. Fixed in feature-flags/rhf/storage.

**Also addressed:**
- `List.Item`/`NestedList.Item`/`Sidebar.Item` now default to `size="md"` instead of `sm` — closer to their old `h-7`/`text-sm` than `sm`'s `h-6`/`text-xs`, per the reviewer's note that the size delta was undersold as "reconciled heights."
- Stripped the redundant explicit `h-screen`/`flex-1` classNames the consumer codemod had added to `PluginShell`/`EmptyState` call sites, now that those are defaults again (harmless via tailwind-merge, but confusing to read).
- Renamed the stale `WithAdornment` story export to `WithTrailing`.
- Updated the changeset to cover all of the above.

**Declined:** the suggestion to split hover out of `surfaceTone` into a fully separate layer rather than a variant flag — the `interactive: true` variant does the same job with less surface area, so I went with that instead of introducing a new abstraction.

All 8 consumer packages individually typecheck/lint clean, and the full monorepo `pnpm lint:all` (62/62) / `pnpm typecheck:all` (63/63) both pass after these fixes.

V3RON:
## Opus review complete — 2 real bugs + 5 should-fix issues, all fixed

An Opus review of the diff against #421 (Phase 2) turned up two genuine bugs and several real-but-lower-severity issues. All are fixed and pushed.

**Bugs:**
- `overlay-plugin/src/ui/styles.css` still referenced the removed `--destructive` var (`--color-danger: var(--destructive)`), leaving `--color-danger` unresolvable in that plugin. This file layers on top of `@rozenite/ui/styles.css`, unlike the plugin's fully-independent `globals.css`, which was correctly left untouched.
- Removing `mt-auto` from `Sidebar.Footer` broke the shell's *collapsed* rail specifically — the expanded state has a `flex-1` scroll spacer that naturally pushes the footer down, but the collapsed state has no such spacer, so the footer floated to the top.

**Should-fix:**
- `surfaceTone`'s hover states (added for Button) leaked into `Badge`, a non-interactive `<span>` — every badge looked clickable on hover. Fixed by making hover opt-in via a new `interactive` variant on `surfaceTone`, used by `Button`/`IconButton` only (`Badge`/`Alert` stay non-interactive).
- feature-flags-plugin's migrated reset `IconButton` was missing `tone="neutral"`, rendering purple instead of neutral, and carried a redundant native `title` alongside `IconButton`'s own tooltip.
- Toolbar icon-only buttons in storage/rhf/feature-flags still sized `w-7` (paired with the old `h-7` default) — no longer square now that the default is `size="sm"` (`h-6`). Fixed to `w-6`.
- `List.Item`/`NestedList.Item`/`Sidebar.Item` defaulted to `size="sm"` (`h-6`, `text-xs`), a real readability regression from the old hardcoded `h-7 text-sm` for every existing call site that didn't opt into a size. Changed the default to `size="md"` (`h-8 text-sm`), which is much closer to the original.

**Declined / already-current by inspection:** the reviewer's note about `PluginShell`/`EmptyState` still hardcoding `h-screen`/`flex-1` was accurate at review-launch time but had already been reverted in a prior commit on this branch (evidence from the consumer codemod showed dropping those defaults silently collapses several panels to zero height) — the redundant explicit classNames the codemod added were cleaned up instead. The `soft`-variant hover gap and leading-icon-doesn't-scale-with-`lg`-size nits were left as-is (no current consumer exercises either).

All fixes are typecheck/lint/test-clean across `@rozenite/ui`, the storybook app, and all 8 adopting consumer packages, plus a full `pnpm -w run typecheck:all` / `lint:all` (63/62 tasks green).

V3RON:
## Independent review + fixes

Since this PR's history includes two duplicate review runs from an orchestration mishap (see below), I spawned a fresh, isolated review agent — with no visibility into the prior review commentary — to re-audit the full diff from scratch, including the 8-consumer-package codemod and the tone/variant-split internals.

**Verdict: no real bugs found.** It confirmed: no leftover old-API references anywhere in the repo, no `variant` used without an accompanying `tone`, correct color equivalence for every migrated combination, no hover-state leakage onto static surfaces (Badge/Alert), correct accessible-name threading on all 11 `Button size="icon"` → `IconButton` migrations, and that the `h-screen`/`flex-1` "keep as default" decision is followed through consistently with no redundant or missing overrides at any consumer call site.

It did surface a few real should-fix issues, now fixed and pushed:
- `surfaceTone`'s `soft` variant had no hover state even when `interactive: true` was passed — `variant="soft"` type-checked on Button/IconButton but rendered non-interactively. Added hover for `soft`, safe now that it's gated behind `interactive` (Alert, the only static `soft` consumer, never opts in).
- `SearchField` and `Combobox.Input` had hardcoded icon/padding/clear-button positioning that only matched the `md` size — `sm`/`lg` produced misaligned icons and clipped padding. Both are now fully size-aware.
- `Badge`'s `sm` step used an arbitrary `text-[10px]` instead of the established `text-xs`, breaking from the rest of the scale.
- The changeset was missing a few real user-visible changes (Toolbar.Button's/PluginHeader.ThemeSwitcher's height, EmptyState's spacing, the now-colorless `buttonVariants()`/`badgeVariants()` exports, and the additive-only changes) — table updated.
- `storage-plugin`'s test mock for `@rozenite/ui` was missing an `IconButton` export after the codemod moved icon-only row actions onto it — latent trap, now fixed.
- Toast's `viewportClassName` JSDoc default string didn't match the real base classes.

**On the duplicate reviews:** two of the parallel migration agents I dispatched for the 8-consumer-package codemod overstepped their scope (each was told to touch exactly one package and not commit anything) and instead independently completed the rest of the phase workflow, including opening this PR and each running its own Opus review — hence the two review comments above this one. I verified the resulting code directly rather than trusting either of those reviews, and the fresh audit above is the one to trust.

All validation re-run and green: `@rozenite/ui` typecheck/lint/build/test (33/33), Storybook typecheck/lint, full monorepo `lint:all` (62/62) and `typecheck:all` (63/63), plus a visual check of the size-aware SearchField/Combobox/Badge fixes in Storybook.

PR #425, opened 2026-08-18
feat(ui): add useConfirmDialog and migrate plugin panels to it

V3RON:
## Description

Adds `useConfirmDialog()` to `@rozenite/ui`: an imperative counterpart to `ConfirmDialog` that resolves a promise with the user's choice instead of driving `open`/`onOpenChange` from local state. `PluginShell` now mounts `Toast` and `ConfirmDialog.Provider` automatically for every plugin panel, so panels no longer need to hand-mount `<Toast.Provider>` to use `useToast()`.

Migrates the feature-flags, React Hook Form, and storage DevTools panels to `useConfirmDialog()`, replacing their declarative confirm/alert dialogs (open/onOpenChange + local state) and dropping the now-redundant `Toast.Provider` wrappers.

## Related Issue

No existing issue tracks this change; opened without one at the requester's direction.

## Context

- `ConfirmDialog`'s existing default export is renamed internally to `ConfirmDialogRoot`; `ConfirmDialog` itself is unchanged as a callable export and gains a `Provider` (`ConfirmDialog.Provider`), mirroring the `Toast`/`Toast.Provider` pattern already in the package.
- `ConfirmDialogProvider` queues at most one pending confirmation at a time; a second `confirm()` call while one is showing waits its turn.
- In `storage-plugin`, delete-entry and purge-storage now capture their target/key in the click-time closure instead of separate confirmation-pending state. This means the confirmation always acts on what the user was looking at when they clicked, even if the selected storage changes while the dialog is open — a behavior change worth flagging for review, though arguably more correct for a devtools panel.
- Added a version plan (`.changeset/imperative-confirm-dialog.md`) covering `@rozenite/ui`, `@rozenite/feature-flags-plugin`, `@rozenite/storage-plugin`, and `@rozenite/rhf-plugin`.
- Added a Storybook story (`ConfirmDialog.stories.tsx` → `Imperative`) demonstrating `useConfirmDialog()` under `ConfirmDialog.Provider`.

## Testing

Automated:

- `pnpm --filter @rozenite/ui typecheck / lint / test`
- `pnpm --filter @rozenite/storage-plugin --filter @rozenite/feature-flags-plugin --filter @rozenite/rhf-plugin typecheck / lint / test`
- `pnpm typecheck:all`, `pnpm lint:all`, `pnpm format:all` (repo-wide, all green aside from pre-existing unrelated formatting drift in 3 untouched files)
- `pnpm release:plan` — confirms the changeset is picked up

Manual:

- Started `apps/ui-storybook` and used `agent-browser` to open the new `Imperative` `ConfirmDialog` story, click through to open the dialog, confirm it, and verify the promise resolved (`true`) and the demo rendered "Confirmed".

### DIFF
--- a/.changeset/imperative-confirm-dialog.md
+++ b/.changeset/imperative-confirm-dialog.md
@@ -1,0 +1,10 @@
+---
+'@rozenite/feature-flags-plugin': minor
+'@rozenite/storage-plugin': minor
+'@rozenite/rhf-plugin': minor
+'@rozenite/ui': minor
+---
+
+Add `useConfirmDialog()` to `@rozenite/ui`: an imperative alternative to `ConfirmDialog` that resolves a promise with the user's choice instead of driving `open` from local state. `PluginShell` now mounts `Toast` and `ConfirmDialog.Provider` automatically, so plugin panels no longer need to wrap themselves in `<Toast.Provider>` to use `useToast()`.
+
+Migrate the feature-flags, storage, and React Hook Form DevTools panels to `useConfirmDialog()`, replacing their declarative confirm/alert dialogs.
--- a/apps/ui-storybook/src/stories/ConfirmDialog.stories.tsx
+++ b/apps/ui-storybook/src/stories/ConfirmDialog.stories.tsx
@@ -1,5 +1,6 @@
 import type { Meta, StoryObj } from '@storybook/react-vite';
-import { Button } from '@rozenite/ui';
+import { useState } from 'react';
+import { Button, useConfirmDialog } from '@rozenite/ui';
 import { ConfirmDialog } from '@rozenite/ui';
 const meta = {
   component: ConfirmDialog,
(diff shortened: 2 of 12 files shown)


V3RON:
Ran an independent review of this PR (via a separate agent). It found one critical bug and several smaller issues; both are fixed in the two commits just pushed.

**Must-fix (fixed):** `useToast()`/`useConfirmDialog()` were called in the component that *renders* `<PluginShell>` (e.g. `FeatureFlagsPanelContent`), which sits above `PluginShell` in the tree — outside the `Toast`/`ConfirmDialog.Provider` it mounts around its own children. This threw immediately for `useConfirmDialog`, and would have thrown for `useToast` too once the old `Toast.Provider` wrappers were removed. Fixed by moving `<PluginShell>` up into each plugin's default export, so the `*PanelContent` component using the hooks is a child of `PluginShell`, not its parent.

**Should-fix (fixed):**
- The storage-plugin `columns` `useMemo` closed over a `handleDeleteClick` that was a new function identity every render, defeating the memo and giving `@tanstack/react-table` a new `columns` array on every keystroke. Wrapped `handleDeleteClick` in `useCallback`.
- A pending `confirm()` promise could hang forever if the `ConfirmDialogProvider` unmounted mid-confirmation. It now resolves `false` on unmount.
- Queued confirmations reused the same `ConfirmDialogRoot` instance across unrelated dialogs (no title/content swap animation, shared internal state). Added `key={pending.id}`.

**Nits (fixed):** unawaited `showAlert()` calls in storage-plugin now use `void` for consistency with the rest of the diff; `ConfirmDialogProviderProps` is now exported from the package barrel; added unit tests for `ConfirmDialogProvider`/`useConfirmDialog` (confirm, cancel, and queueing), which would have caught the critical bug above.

**Not changed (flagged, left as-is / follow-up):**
- Purge/delete confirmations no longer close automatically when the selected storage target changes mid-dialog (previously they did); they now act on the target/key captured at click time instead. This is called out in the PR description as an intentional, arguably-more-correct behavior change for a devtools panel, but there's no cancel affordance if that's ever wanted.
- The `unstyled` `PluginShell` branch (documented as "skip design tokens entirely") now also mounts `Toast`/`ConfirmDialog.Provider`, whose markup depends on design tokens. Matches the original plan but is a minor inconsistency with that branch's stated purpose.

Full validation re-run after the fixes: `pnpm typecheck:all`, `pnpm lint:all`, `pnpm release:plan`, and the affected packages' test suites — all green.

PR #448, opened 2026-08-21
feat(middleware): resolve per-target CDP capability profiles for agent sessions

### LINKED ISSUE #447: Rozenite for Agents advertises built-in domains that a Lynx target cannot back
## Summary

Rozenite for Agents exposes five built-in domains — `console`, `network`, `react`, `performance`, `memory` — and advertises all five on every session regardless of what the connected target is. Those domains were designed against React Native's CDP surface. Lynx (added in #443) exposes a different one, so on a Lynx target several of them cannot work, and there is no way for an agent to discover that except by calling and failing.

## What Lynx actually exposes

Verified against `lynx-family/lynx` (`LynxDevToolNG::RegisterAgents`, the list of CDP domains a Lynx device answers at all) and `lynx-family/primjs` (`src/inspector/protocols.cc`, the method table behind the domains Lynx forwards into the JS engine):

- **No `Network` domain is registered at all.** `Network.enable` returns `Not implemented` (`-32601`), so the whole `network` domain is dead.
- **No React DevTools backend exists**, and `@rozenite/lynx`'s dispatcher only initialises the `rozenite` domain — so `react-devtools` messages are never delivered and the `react` domain returns nothing.
- **`Tracing` exists but is Perfetto-based, not Chrome-based.** It ignores Chrome trace categories and reports completion with an `IO` stream handle rather than `Tracing.dataCollected` events.
- **`HeapProfiler` is forwarded to PrimJS, which implements `takeHeapSnapshot` but not sampling.**
- `console` works: PrimJS emits `Runtime.consoleAPICalled` and Lynx's own `Log` agent emits `Log.entryAdded`. Plugin and app-defined domains work too, since they travel over Rozenite's own device channel.

## Why this is worse than an error

Unsupported is not one behaviour on the wire but three, and two of them are silent:

1. **Unregistered domain** → `-32601`. Loud, but the agent only learns after spending a turn, and the rejection today is a stringified error object that does not even name the method.
2. **Unknown method in an engine-backed domain** → PrimJS answers with an empty success. The call *resolves*; the event the caller awaits never arrives. `stopSampling` returns nothing, and nothing bounds the wait.
3. **Same name, different protocol** → `Tracing` succeeds and reports completion over a stream Rozenite never reads, so `stopTrace` finalises a syntactically valid trace artifact containing zero events and reports success.

Cases 2 and 3 cannot be detected by inspecting a response. They can only be prevented by knowing, before the call, that this target does not support this tool.

## Proposal

Resolve a **capability profile** per session from its target's platform, and use it at three widening radii: don't register unsupported tools, publish the gaps alongside the tool list so `rozenite agent domains` reports an availability and a reason, and fail at resolution with an actionable message (`network` should name `@rozenite/network-activity-plugin` as its fallback) rather than a protocol error.

Frame it as a capability *descriptor* rather than a Lynx blocklist: the same mechanism could later add Lynx-only domains (`UITree`, `Template`, Lynx's own `Performance`/`Memory`) or describe the next platform.

Separately, three fixes to the CDP command channel are platform-agnostic and worth having regardless: name the method in device errors, bound waits on device events, and refuse to finalise an empty trace artifact. The trace hang in particular affects React Native too — a device that dies mid-trace hangs `stopTrace` forever today.

V3RON:
## Description

Rozenite for Agents now knows which built-in domains the connected target can actually back, instead of advertising all five on every session.

- Each session resolves a **capability profile** from its target's platform. Unsupported tools are never registered, so `list-tools` is honest.
- Unavailable domains stay visible in `rozenite agent domains`, which gains an `availability` column (`supported` / `degraded` / `unsupported`) and a `fallback` — both default fields, not `--verbose` ones.
- Calling an unsupported tool now fails during resolution with the reason and a next step, before anything reaches the wire:

  ```
  Tool "listRequests" is not supported on this lynx target. Lynx registers no CDP
  Network domain, so every Network.* command returns "Not implemented" (-32601).
  Use the `@rozenite/network-activity-plugin` domain instead.
  ```

- On a Lynx target that means `console`, plugin domains, app tools and `memory.takeHeapSnapshot` are supported; `memory` is degraded (allocation sampling removed); `network`, `react` and `performance` are unavailable with reasons.

Three fixes to the CDP command channel apply to **every** platform:

- A device error names the method it refused (`UnsupportedCdpMethodError` for `-32601`) instead of arriving as a stringified error object.
- Waits on a device event are bounded, so a capture whose completion event never arrives fails with a diagnosis rather than hanging forever.
- `stopTrace` refuses to finalise a zero-event trace artifact instead of reporting a successful capture of nothing.

Agent docs gain a `platforms` frontmatter field, platform-availability sections on `network`, `react`, `performance` and `memory`, and a new `lynx` orientation doc.

## Related Issue

Closes https://github.com/callstackincubator/rozenite/issues/447

## Context

The five built-in domains were designed against React Native's CDP surface. Lynx's is different, verified against `LynxDevToolNG::RegisterAgents` (the domains a Lynx device answers at all) and `primjs`'s `src/inspector/protocols.cc` (the method table behind the domains Lynx forwards into the JS engine): no `Network` domain is registered, no React DevTools backend exists, `Tracing` is Perfetto-based rather than Chrome-based, and `HeapProfiler` implements snapshots but not sampling.

The design follows from the fact that **"unsupported" is three behaviours on the wire, and only one is loud**: an unregistered domain answers `-32601`; an unknown method inside an engine-backed domain answers with an *empty success*, so the call resolves and the awaited event never comes; and a same-named-different-protocol domain succeeds while producing the wrong thing. The latter two cannot be detected from a response, which is why this is resolved from a profile before the call rather than translated from an error after it.

Some decisions worth a reviewer's attention:

- **The table is static, not probed.** A probe costs round trips at session start and cannot see the empty-success case anyway. The one axis that genuinely varies at runtime is the V8-vs-PrimJS backend, which already has a free signal (only V8 emits a real `Runtime.executionContextCreated`); that is left as a future refinement rather than a foundation.
- **Absence means supported.** React Native's profile is empty and its behaviour is unchanged; nobody maintains an allowlist that grows with every new tool.
- **Unavailable domains are still listed.** Dropping `network` outright would make `rozenite agent network listRequests` fail with `Unknown domain "network". Did you mean…?`, which reads to an agent as a typo to correct rather than a platform limit to route around.
- **Both new response fields are optional.** An older CLI against a newer server ignores them; a newer CLI against an older server reads `undefined` and treats every domain as supported — exactly the previous behaviour. No lockstep upgrade.
- **Filtering is a wrapper, not a branch per service.** The four domain services stay ignorant of platforms; `withCapabilityFilter` returns the original service untouched when nothing is filtered. `callTool` is guarded as well as `getTools`, since the call-tool route reaches tools by exact name.
- **Platform comes from two places.** The dev server's `platform` config is the default; `@rozenite/lynx-dev` additionally stamps `rozenite: { platform: 'lynx' }` on each `/json/list` page, which `metro-discovery` carries onto `MetroTarget` and which wins per target. Metro never emits that key and every other consumer ignores unknown ones, so it is additive.

Two gaps I'd rather name than leave implicit:

- The profile drift guard covers `memory` and `network` — the domains whose services construct cheaply in a test. `react` and `performance` have no tool-level entries today so nothing is currently unguarded, but adding one there would not be caught.
- A Lynx-only domain over `UITree` / `Template` / Lynx's own `Performance` and `Memory` is now a profile entry away, but it is new surface area rather than a fix and stayed out of scope.

## Testing

Automated, from the repository root after `git fetch origin main`:

- `pnpm checks:affected` — typecheck, lint and format across 102 tasks, all passing.
- `pnpm test:affected` — 63/63 tasks passing.
- `pnpm release:plan` — confirms the version plan.

New tests in `packages/middleware/src/__tests__/agent-capabilities.test.ts` cover the React Native profile declaring no gaps, a drift guard asserting the Lynx table names only tools the services actually expose, whole-domain versus tool-level gap reporting, filtering on both `getTools` and `callTool`, and the untouched-service fast path.

Three existing tests changed deliberately: the two `agent-command-output` domain-listing assertions carry the new default columns (their mocks now include a `degraded` and an `unsupported` domain so the columns are exercised), and `skills-registry` counts the new `lynx` doc.

Not manually verified against a physical Lynx device — the Lynx behaviour encoded in the profile is sourced from the upstream C++ registration and dispatch tables cited above rather than from a live session.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Pj8YXsJh4mMsJ1HpJEr7B5)_

### DIFF
--- a/.changeset/lynx-agent-capabilities.md
+++ b/.changeset/lynx-agent-capabilities.md
@@ -1,0 +1,41 @@
+---
+'@rozenite/agent-shared': minor
+'@rozenite/middleware': minor
+'@rozenite/agent-sdk': minor
+'@rozenite/lynx-dev': minor
+'rozenite': minor
+---
+
+Teach Rozenite for Agents which built-in domains the connected target can
+actually back, so a Lynx session no longer looks like a React Native one.
+
+The five built-in domains were designed against React Native's CDP surface, and
+Lynx exposes a different one: it registers no `Network` domain at all, has no
+React DevTools backend, and implements a Perfetto-based `Tracing` domain that
+shares Chrome's method names but not its protocol. Until now every session
+advertised all five regardless, so an agent on a Lynx target could only find out
+by calling — and two of the three ways that fails are silent. `network` errored
+with a raw protocol code, `react` returned nothing, `performance` finalised a
+trace artifact containing zero events, and heap sampling appeared to start and
+collected nothing.
+
+Each session now resolves a capability profile from its target's platform.
+Unsupported tools are never registered, so `list-tools` is honest; unavailable
+domains stay visible in `rozenite agent domains` with an `availability` column, a
+reason, and — for `network` — the `@rozenite/network-activity-plugin` domain to
+use instead. Calling an unsupported tool fails during resolution with that same
+explanation rather than a protocol error, so the agent gets a next step instead
+of a dead end. On Lynx that means `console`, plugin domains, app tools and
+`memory.takeHeapSnapshot` are reported supported, `memory` is degraded, and
+`network`, `react` and `performance` are unavailable with reasons.
+
+Three fixes to the CDP command channel apply to every platform, not just Lynx. A
+device error now names the method it refused instead of arriving as a stringified
+error object; waits on a device event are bounded, so a capture whose completion
+event never arrives fails with a diagnosis instead of hanging forever; and
+`stopTrace` refuses to hand back an empty trace artifact rather than reporting a
+successful capture of nothing.
+
+Both new fields on the session tools response are optional, so a CLI and a Metro
+on different versions keep working together — an older server simply reports no
+capability data and every domain is treated as supported, exactly as before.
--- a/.changeset/lynx-agent-capabilities.md
+++ b/.changeset/lynx-agent-capabilities.md
@@ -2,7 +2,6 @@
 '@rozenite/agent-shared': minor
 '@rozenite/middleware': minor
 '@rozenite/agent-sdk': minor
-'@rozenite/lynx-dev': minor
 'rozenite': minor
 ---
 
@@ -19,7 +18,8 @@ with a raw protocol code, `react` returned nothing, `performance` finalised a
 trace artifact containing zero events, and heap sampling appeared to start and
 collected nothing.
 
-Each session now resolves a capability profile from its target's platform.
+Each session now resolves a capability profile from the integration its dev
+server hosts.
 Unsupported tools are never registered, so `list-tools` is honest; unavailable
 domains stay visible in `rozenite agent domains` with an `availability` column, a
 reason, and — for `network` — the `@rozenite/network-activity-plugin` domain to
@@ -29,12 +29,13 @@ of a dead end. On Lynx that means `console`, plugin domains, app tools and
 `memory.takeHeapSnapshot` are reported supported, `memory` is degraded, and
 `network`, `react` and `performance` are unavailable with reasons.
 
-Three fixes to the CDP command channel apply to every platform, not just Lynx. A
+Three fixes to the CDP command channel apply to every integration, not just
+Lynx. A
 device error now names the method it refused instead of arriving as a stringified
 error object; waits on a device event are bounded, so a capture whose completion
 event never arrives fails with a diagnosis instead of hanging forever; and
-`stopTrace` refuses to hand back an empty trace artifact rather than reporting a
-successful capture of nothing.
+`stopTrace` refuses to hand back a trace artifact containing no events rather
+than reporting a successful capture of nothing.
 
 Both new fields on the session tools response are optional, so a CLI and a Metro
 on different versions keep working together — an older server simply reports no
--- a/packages/agent-sdk/package.json
+++ b/packages/agent-sdk/package.json
@@ -51,6 +51,7 @@
   },
   "dependencies": {
     "@rozenite/agent-shared": "workspace:*",
+    "@rozenite/tools": "workspace:*",
     "tslib": "^2.3.0"
   },
   "devDependencies": {
--- a/packages/agent-sdk/src/__tests__/domain-utils.test.ts
+++ b/packages/agent-sdk/src/__tests__/domain-utils.test.ts
@@ -3,9 +3,12 @@ import { fileURLToPath } from 'node:url';
 import { describe, expect, it } from 'vitest';
 import type { AgentTool } from '@rozenite/agent-shared';
 import {
+  applyCapabilities,
   buildRuntimePluginDomains,
   deriveDomainName,
   formatUnknownDomainError,
+  formatUnsupportedToolError,
+  isToolUnavailable,
   getDomainToolsByDefinition,
   inferPluginId,
   inferToolShortName,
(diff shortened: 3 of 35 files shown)


V3RON:
Rebased onto `main` and reworked in response to review. The branch was cut before #456/#457 landed, and the collision with them was more than textual.

**Integration, not platform.** `AgentTargetPlatform` was the same union as `RozeniteHostIntegration` under the name #456 deliberately rejected — `platform` means the device OS everywhere else in this system. The agent surface now sources `RozeniteIntegration` from `@rozenite/tools/integration`; the session field, the wire field and `CapabilityProfile` are all `integration`. `packages/middleware/src/config.ts` is untouched again — `RozenitePlatform` was reinstating a type main had just replaced.

**One signal for "is this target Lynx", not two.** The `/json/list` `rozenite.platform` extension and `MetroTarget.platform` are gone. `@rozenite/lynx-dev` already sets `integration: 'lynx'` on the middleware config, so the two could never disagree — the per-target override bought nothing and duplicated the mechanism #456 established. The host integration alone resolves the profile.

**`lynx-web` answered explicitly.** `resolveCapabilityProfile` now takes the full four-id union. A web target is a browser and backs Chrome's CDP surface in full, so both `-web` variants get the empty profile rather than `lynx-web` falling through to PrimJS's gaps.

Four review findings fixed:

- **The empty-trace guard checked the wrong thing.** It tested whether the header had been written, but the header was written per event *batch*, before the loop. A `Tracing.dataCollected` whose `value` is absent or not an array yields zero events, still opened the artifact, and passed the guard — exactly what a Tracing domain that only shares Chrome's name would send. It now counts events written, and the header is written lazily per event. Covered by three tests in `agent-local-domains.test.ts`; the middle one was confirmed to fail against the previous implementation.
- **Tool-level gaps never reached the client's error path.** `memory` is degraded on Lynx, so `startSampling` was filtered out server-side and the agent got the generic unknown-tool error — a typo-shaped answer for an integration limit, one level below the case this mechanism exists for. `UnsupportedDomainInfo.tools` now carries `{ name, reason }` per tool, `applyCapabilities` folds that onto the domain, and `isToolUnavailable` gates resolution. Per-tool reasons also remove the domain borrowing the alphabetically-first tool's reason, which would have explained the wrong thing the first time a domain had two unrelated gaps.
- **The `platforms` frontmatter was dead.** Parsed into `SkillDoc` and printed by nothing. Renamed to `integrations` (these are integration ids, not device OSes) and emitted by `rozenite skills list`.
- **`AgentSessionInfo.platform` was required** while the tools-response field was optional, contradicting the stated cross-version story. Now optional.

Also: `withCapabilityFilter` snapshots `getTools()` once instead of calling it three times, and `DomainListRow.availability` is typed `ToolAvailability` rather than `string`.

New tests: SDK-side capability folding in `domain-utils.test.ts` (none existed), the three trace-guard tests, per-tool reasons and the `-web` profile decision in `agent-capabilities.test.ts`, and the `integrations` frontmatter in `skills-registry.test.ts`.

Validation from the repository root after `git fetch origin main`: `pnpm checks:affected` and `pnpm test:affected` both pass. Still not verified against a physical Lynx device — the profile remains sourced from the upstream C++ tables cited in the description.

PR #464, opened 2026-08-24
feat(middleware): expose complete React commit timeline

### LINKED ISSUE #462: Expose the complete React profiler commit timeline from stopProfiling
### Description

The React agent `stopProfiling` tool currently returns a duration-ranked, size-limited `topSlowCommits` list. The underlying React DevTools profiling data contains every commit, but callers cannot inspect the full chronological timeline without already knowing each root and commit index. This limits downstream analysis that needs to correlate render bursts, effects, priorities, updater counts, and change descriptions across the entire recording.

### Suggested solution

Add an untruncated `commits` array to the `stopProfiling` result. Each item should identify the root and commit index and include the existing React DevTools metadata needed for post-processing:

- render, effect, and passive-effect duration
- timestamp and priority
- rendered-fiber and updater counts
- whether change descriptions are available

Sort the array chronologically, with deterministic root/commit tie-breakers. Keep `topSlowCommits` unchanged for backward compatibility and quick scouting.

I can submit the implementation and regression tests.

avocado-cb:
## Description

Adds an additive `commits` field to the React agent `stopProfiling` result. The field contains every captured commit in chronological order with:

- root and commit identifiers
- render, effect, and passive-effect durations
- timestamp and priority
- rendered-fiber and updater counts
- change-description availability

The existing duration-ranked, size-limited `topSlowCommits` field remains unchanged for backward compatibility and quick scouting. The React agent documentation and middleware release plan are updated alongside the implementation.

## Related Issue

Closes https://github.com/callstackincubator/rozenite/issues/462

## Context

`stopProfiling` currently exposes only the slowest commits, even though the normalized React DevTools profile retains the full recording. Downstream performance analysis needs the complete timeline to correlate render bursts and post-process the same commit metadata available to profiler tooling.

The timeline is sorted by timestamp, then root id and commit index for deterministic ordering across roots. A regression test covers ordering, metadata, and preservation of the existing slow-commit summary.

## Testing

- `pnpm checks:affected`
- `pnpm test:affected`
- `pnpm release:plan`
- Pre-push repository-wide lint and typecheck hooks

### DIFF
--- a/.changeset/react-commit-timeline.md
+++ b/.changeset/react-commit-timeline.md
@@ -1,0 +1,5 @@
+---
+'@rozenite/middleware': minor
+---
+
+Expose every captured React profiler commit from `stopProfiling` in chronological order, including render and effect timing, priority, rendered-fiber and updater counts, and change-description availability. The existing slow-commit summary remains unchanged for quick scouting.
--- a/packages/cli/docs/react.md
+++ b/packages/cli/docs/react.md
@@ -29,6 +29,12 @@ Search and inspect:
 Profile:
 `startProfiling` -> reproduce interaction -> `stopProfiling` -> `getRenderData`.
 
+`stopProfiling` returns every captured commit in an untruncated, chronological
+`commits` array with render/effect timing, priority, rendered-fiber and updater
+counts, and change-description availability. The duration-ranked,
+size-limited `topSlowCommits` list remains available for quickly finding the
+slowest commits before calling `getRenderData`.
+
 `getRenderData` rows carry `displayName` and timing by default. Add
 `--fields ...,changedKeys` for the exact changed prop, state, and context key
 names behind `changeTypeHints`, which attributes a re-render without a
--- a/packages/middleware/src/agent/runtime/react/__tests__/store.test.ts
+++ b/packages/middleware/src/agent/runtime/react/__tests__/store.test.ts
@@ -511,11 +511,22 @@ describe('React tree store getComponent', () => {
 
 type MockPhase = 'idle' | 'profiling' | 'processing';
 
+type MockCommitData = {
+  duration: number;
+  timestamp: number;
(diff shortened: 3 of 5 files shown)


V3RON:
Thanks for this, and sorry to close it — the underlying need was real and we've shipped it, just through a different surface.

#478 (merged) adds the missing commit metadata to the existing `getProfileTimeline` tool as opt-in pagination fields: `effectDurationMs`, `passiveEffectDurationMs`, `priorityLevel`, `updaterCount` and `hasChangeDescriptions`. Those were the five fields this PR carried that the timeline tool did not, and as you noted they were already normalized and retained by `profiling-store.ts` — only unexposed.

The part we went a different way on is delivery. `getProfileTimeline` already covers what #462 asked for — every commit, chronologically, with deterministic root and commit tie-breakers — and additionally paginates with a cursor, filters by `rootId`, sorts `timeline` or `duration-desc`, and stays queryable after `stopProfiling` has returned. An untruncated `commits[]` array inline in a tool result is the shape cursor pagination exists to avoid, and it would have given us two overlapping surfaces for the same data.

`defaultFields` is unchanged in #478, so default responses don't grow; the new fields come back via `--fields` or `--verbose`. Verified against the playground on a simulator:

```
cols: [rootId, commitIndex, durationMs, effectDurationMs, passiveEffectDurationMs, priorityLevel, updaterCount, hasChangeDescriptions]
rows: [[1,0,34.59,1.775,0.953,"Immediate",1,true], [1,1,1.53,0.231,0.015,"Immediate",1,true], ...]
```

#462 is closed by #478.

Your framing in #462 is what drove the field list, and the effect-duration point in particular turned out to matter more than we'd have guessed — a commit that renders in 34.59ms while spending another 1.775ms in layout effects is invisible on render duration alone. If you're up for another one, #479 and #480 came out of the same audit and are both open.

PR #465, opened 2026-08-25
fix(middleware): negotiate loopback debugger origins

### LINKED ISSUE #463: Agent session fails for numeric loopback debugger origins
## Description

`@rozenite/middleware` derives the debugger WebSocket `Origin` directly from the inspector target URL. When the target uses a numeric loopback host such as `ws://127.0.0.1:<port>`, some React Native Metro/dev-middleware configurations reject the resulting `Origin: http://127.0.0.1:<port>` with HTTP 401 while accepting the equivalent `Origin: http://localhost:<port>`.

That prevents `agent session create` from completing the CDP bootstrap even though target discovery and the debugger URL itself are reachable.

## Reproduction

1. Run a React Native app whose inspector target is exposed through numeric loopback, for example `ws://127.0.0.1:8081/inspector/debug?...`.
2. Create a Rozenite agent session for that target.
3. Observe the debugger WebSocket upgrade fail with HTTP 401 because the outgoing Origin uses the numeric loopback host.
4. Retry with the same debugger URL but `Origin: http://localhost:8081`; the upgrade succeeds.

## Expected behavior

Canonicalize IPv4 and IPv6 loopback hosts to `localhost` when constructing the debugger WebSocket Origin:

- `ws://127.0.0.1:8081/...` → `http://localhost:8081`
- `ws://[::1]:8081/...` → `http://localhost:8081`

Non-loopback hosts and the `ws:` / `wss:` protocol mapping should remain unchanged.

## Context

This is a follow-up edge case to #296, #297, and #307. The debugger URL should remain untouched; only the outbound Origin header needs loopback canonicalization.

avocado-cb:
## Description

Negotiate the debugger WebSocket Origin for loopback Metro sessions instead of unconditionally rewriting numeric loopback addresses to `localhost`.

- Try the Origin matching `webSocketDebuggerUrl` first.
- Retry loopback connections with the alternate `localhost` or `127.0.0.1` Origin.
- Keep retry eligibility open until the initial `ReactNativeApplication.enable` and `Runtime.enable` CDP commands succeed, because an Expo Origin mismatch is terminated after the WebSocket upgrade.
- Preserve remote and secure debugger Origins without fallback.

## Related Issue

Closes https://github.com/callstackincubator/rozenite/issues/463

## Context

The React Native and Expo layers validate different stages of this connection:

- React Native dev-middleware requires an Origin during the WebSocket upgrade and rejects a missing header with HTTP 401. It accepts common loopback hostnames.
- Expo CLI performs a second exact-host check after the upgrade. If the Origin host differs from Expo's configured `serverBaseUrl`, Expo terminates the socket and the client sees close code 1006 even though the WebSocket emitted `open`.

Expo's configured hostname is environment-dependent. `REACT_NATIVE_PACKAGER_HOSTNAME=localhost` makes it expect `localhost`; with that variable unset, Expo's URL creator normalizes its localhost host type to `127.0.0.1`. Meanwhile, `/json/list` builds `webSocketDebuggerUrl` from the discovery request's Host header. The target URL therefore does not always identify the Origin Expo will accept.

That explains why a localhost-only rewrite can fix one Metro process while breaking another. The fallback must work in both directions and must cover post-upgrade termination, not only HTTP handshake rejection.

## Testing

- `pnpm release:plan`
- `pnpm checks:affected`
- `pnpm test:affected` — 173 middleware tests passed; all affected tasks passed
- `pnpm --config.verify-deps-before-run=warn --filter @rozenite/middleware exec vitest --run src/__tests__/agent-session.test.ts` — 31 tests passed
- Started the playground Metro server; all 14 plugins loaded and Metro reached `Waiting on http://localhost:8082`
- Registered a synthetic inspector page and verified both Expo configurations:
  - `REACT_NATIVE_PACKAGER_HOSTNAME=localhost`: numeric Origin upgraded and then closed with 1006; localhost remained connected.
  - variable unset: numeric Origin remained connected; localhost upgraded and then closed with 1006.
  - missing Origin was rejected with HTTP 401.

### DIFF
--- a/.changeset/normalize-loopback-debugger-origin.md
+++ b/.changeset/normalize-loopback-debugger-origin.md
@@ -1,0 +1,7 @@
+---
+'@rozenite/middleware': patch
+---
+
+Try the debugger URL's matching WebSocket Origin first, then retry loopback
+connections with the alternate localhost or 127.0.0.1 Origin when Metro rejects
+or terminates the initial connection.
--- a/packages/middleware/src/__tests__/agent-session.test.ts
+++ b/packages/middleware/src/__tests__/agent-session.test.ts
@@ -360,7 +360,7 @@ describe('agent session', () => {
     });
   });
 
-  it('preserves the debugger websocket host in the origin header', () => {
+  it('preserves an IPv4 loopback debugger host in the primary origin header', () => {
     const target = createTarget({
       webSocketDebuggerUrl: 'ws://127.0.0.1:8081/debug',
     });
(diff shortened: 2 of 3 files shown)


V3RON:
Thanks for contribution. While going through your implementation an alternative solution came to my mind, which doesn't require any retries or hardcoding URLs.

https://github.com/callstackincubator/rozenite/pull/468

